PR #3, opened 2020-10-20
Test commit.

DavidSpickett:
see if azure is setup right


DavidSpickett:
Azure seems to be working as expected now.

PR #5, opened 2020-10-26
Skip the condition when read() returns 0 in shell demo

### LINKED ISSUE #4: Fail to run shell demo
**Describe the bug**
The shell demo fails to open stdin to read the input from terminal.

**Configuration**
 - cmake version (`cmake --version`)
```
cmake version 3.16.3
```
 - toolchain/compiler version (`gcc --version`, for example)
```
aarch64-none-elf-gcc (GNU Toolchain for the A-profile Architecture 9.2-2019.12 (arm-9.10)) 9.2.1 20191025
arm-none-eabi-gcc (GNU Arm Embedded Toolchain 9-2020-q2-update) 9.3.1 20200408 (release)
```
 - QEMU version (`qemu-system-arm --version`)
```
QEMU emulator version 5.1.0
QEMU emulator version 4.2.1
```

**To Reproduce (complete those that apply):**
- cmake configuration
```
cmake . -DBUILD_PLATFORM=arm
cmake . -DBUILD_PLATFORM=aarch64
cmake . -DBUILD_PLATFORM=thumb
```
- make command
```
make
```
- run command
```
make run_shell
```

**Expected behavior**

The shell demo should work properly. 

**Affected platforms/settings**
List any other settings/platforms/software versions you have tried and whether that had any effect.
e.g. "This issue presents at -O2 but not -O1."

This problem occurs all platflaform listed below
- `arm` 
- `aarch64`
- `thumb`

The error output is like this:

```shell
$ make run_shell
[ 57%] Built target shell
[ 71%] Built target ps
[ 85%] Built target echo
[100%] Built target ls
---------------------
----- AMT Shell -----
---------------------
$ /home/oscar/ARMMultiTasking/demos/shell/shell.c:236 (command_loop) Expected: (got != 0) && "stdin closed!"
make[3]: *** [CMakeFiles/run_shell.dir/build.make:57: run_shell] Error 1
make[2]: *** [CMakeFiles/Makefile2:894: CMakeFiles/run_shell.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:901: CMakeFiles/run_shell.dir/rule] Error 2
make: *** [Makefile:430: run_shell] Error 2
``` 

The problem may occurs in both version mentioned above, but in QEMU `2.11.2`, the shell demo works properly.
And I try to checkout to some previous commit, but it ends up to have some problems when I try to enter something

```shell
$ make run_shell
[ 57%] Built target shell
[ 71%] Built target ps
[ 85%] Built target echo
[100%] Built target ls
---------------------
----- AMT Shell -----
---------------------
$ d?�Um?�Uf?�UUnhandled escape sequence!
make[3]: *** [CMakeFiles/run_shell.dir/build.make:57: run_shell] Error 1
make[2]: *** [CMakeFiles/Makefile2:894: CMakeFiles/run_shell.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:901: CMakeFiles/run_shell.dir/rule] Error 2
make: *** [Makefile:430: run_shell] Error 2
``` 

It seems that QEMU had disabled some settings which is used to be turned on in `2.*` versions but I fail to find the proper setting of it.

Is there any clue to find or fix it?

OscarShiang:
Try to fix #4.

According to the description of `SYS_READ` of semihosting for aarch32
and aarch64:

   * If the return value is 0, the buffer had been filled successfully
   * If the return value equals to the 3rd argument of `read()`, it
      means no bytes were read.

Since we only need to deal with the condition when the return value is 
smaller than the 3rd argument we passed, we can skip the condition when
the above situations occur.

This commit had been tested on the targets: 
* `arm`
* `aarch64`
* `thumb`

compiled by the tool chains below 

```
aarch64-none-elf-gcc (GNU Toolchain for the A-profile Architecture 9.2-2019.12 (arm-9.10)) 9.2.1 20191025
arm-none-eabi-gcc (GNU Arm Embedded Toolchain 9-2020-q2-update) 9.3.1 20200408 (release)
```

and can work properly in both QEMU 5.1 and QEMU 4.2.1

### DIFF
--- a/demos/shell/shell.c
+++ b/demos/shell/shell.c
@@ -233,10 +233,10 @@ static void command_loop(int input) {
   while (1) {
     // -1 for null terminator space
     ssize_t got = read(input, &state.in, INPUT_BUFFER_SIZE - 1);
-    assert((got != 0) && "stdin closed!");
 
-    // got = 1 + size of read seems to mean nothing read
-    if (got == INPUT_BUFFER_SIZE) {
+    // in qemu 2, got = 1 + size of read seems to mean nothing read
+    // qemu version >=4 returns 0 when nothing is read
+    if (!got || got == INPUT_BUFFER_SIZE) {
       continue;
     }
 


demos/shell/shell.c
```diff
@@ -233,10 +233,10 @@ static void command_loop(int input) {
   while (1) {
     // -1 for null terminator space
     ssize_t got = read(input, &state.in, INPUT_BUFFER_SIZE - 1);
-    assert((got != 0) && "stdin closed!");
 
     // got = 1 + size of read seems to mean nothing read
```
DavidSpickett: Mention qemu 2 here

demos/shell/shell.c
```diff
@@ -233,10 +233,10 @@ static void command_loop(int input) {
...
-    if (got == INPUT_BUFFER_SIZE) {
+    // qemu version >= 4 tend to got 0 in the begin
```
DavidSpickett: Reads better as:
// qemu version >=4 returns 0 when nothing is read

DavidSpickett (changes requested):

DavidSpickett (commented):
Also modify the commit subject to reference the shell demo. E.g.
"In shell demo ignore when read() returns 0"

OscarShiang:
The commit had been updated.

DavidSpickett:
Tested locally with QEMU 2.1, looks good to me, thanks!

PR #6, opened 2020-11-08
Create README.md in demos/yielding

OscarShiang:
Add some description of how the yielding mechanism works.

### DIFF
--- a/demos/yielding/README.md
+++ b/demos/yielding/README.md
@@ -1,0 +1,45 @@
+# Yielding
+
+In this demo we show the usage of yielding, which occurs in a computer program during multithreading.
+
+By calling syscall `yield`, the thread gives away the utilization of the CPU to another thread, or a specific thread (discussed in another demo).
+
+To describe thread yielding, we need to introduce:
+
+* `add_thread_from_worker` (in `src/user/thread.c`)
+* `do_scheduler` (in `src/kernel/thread.c`)
+* `yield` (in `src/user/thread.c`)
+
+## `add_thread_from_worker`
+
+When a thread calls `add_thread_from_worker`, the kernel will create a new thread with the given arguments, initialize the permissions, and set its starting point to be the given function (which is `thread_worker()` in this demo).
+
+After the creation of the thread, it will wait to be chosen by the scheduler so it can utilize the CPU.
+
+## `do_scheduler`
+
+The job of the scheduler is to choose the next thread to make use of the CPU.
+
+In AMT we store all of the threads in an array called `all_threads`, so the scheduler simply walks through this array to find the next thread.
+
+Then the current thread will be suspended until the thread has been chosen to run again.
+
+## `yield`
+
+To yield the utilization, the thread calls the syscall `yield` and sets the kind of yielding as `YIELD_ANY`. The scheduler will choose the next thread to run.
+
+This means that the current thread will have to wait until it is yielded to, to run again.
+
+## Walkthrough
+
+At the beginning of the demo we use `set_kernel_config` to enable scheduler logging.
+
+Then we use `add_thread_from_worker` to create a new thread running `thread_worker`.
+
+When the initialization is complete, the first thread (`setup`) starts to run into an endless loop and repeatedly yields to others.
+
+Since there are only two threads, the scheduler will directly choose the thread running `thread_worker` to execute.
+
+After the thread (`thread_worker`) gets the use of the CPU, it counts the number of times it has been yielded to and yields the use. If it has been yielded to twice, it calls `exit()` to end the demo.
+
+This shows that threads are resumed from the point they yield from, if the thread didn't go back to the point it yielded from or restore all its local variables then the demo would effectively become an infinite loop.


demos/yielding/README.md
```diff
@@ -0,0 +1,112 @@
+# Yielding
+
+In this demo, we try to show the usage of yielding, which is one characteristic of a multitasking system. 
+
+By calling syscall `yield()`, the thread gives away the utilization of a CPU to another thread, or a specific thread (will be discussed later in another demo).
```
DavidSpickett: "Using syscall `yield`"

Leave the () off of yield, since the syscall is just yield, yield() is a user space call. (nit pick but might as well)

Since we're single core, "of a CPU" should be "of the CPU".

It would be good to reference the other demo (I think exyielding) but for now "(discussed in another demo)" would be better. (we can add links later)

demos/yielding/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+To achieve thread yielding, we need to introduce the following functions:
```
DavidSpickett: "To describe thread yielding, we need to introduce"

Makes more sense to me. These three are the key functions but not an exhaustive list of what's involved along the way.

demos/yielding/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+* `add_thread_from_worker` (in `src/user/thread.c`)
+* `do_scheduler` (in `src/kernel/thread.c`)
+* `yield` (in `src/user/thread.c`)
```
DavidSpickett, line 11: I wonder if these can be links to the files in the repo, but leave that out for now.
(I think they'd need to be markdown links to do that)

demos/yielding/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+## `add_thread_from_worker`
+
+In AMT, we try to maintain all of the threads in an array called `all_threads`.
```
DavidSpickett: I definitely agree on using AMT as the abbreviation.

Though it should be:
"In AMT we store all of the threads"

There's no "try" here, if we don't then that's a bug :)

demos/yielding/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+By calling this function, the AMT will create a new thread instance with the given arguments, initialize the permissions, and put the function in.
```
DavidSpickett: The start is a bit odd, what about:
"When a thread calls this function, the kernel will..."

When referring to AMT, either this should be "AMT will" or "the AMT kernel will" or "the kernel will". In this case I would go with "the kernel will", just to save some words. (you're going to be saying this a lot and no one will confuse it with "the kernel" aka the Linux kernel)

"a new thread instance"

This can just be "a new thread". (it's not wrong, it is an instance but it's just not needed)

"and put the function in"

I know what you're trying to say here but it isn't clear if you don't know the system. "and set it's starting point to be the given function".
(perhaps showing the function signature would help here?)

Side note: saying "starting point" isn't quite true since it goes thread_start->user function but I think it fits fine here. From the user space point of view that's the starting point.

demos/yielding/README.md
```diff
@@ -0,0 +1,112 @@
...
+* `do_scheduler` (in `src/kernel/thread.c`)
+* `yield` (in `src/user/thread.c`)
+
+## `add_thread_from_worker`
+
+In AMT, we try to maintain all of the threads in an array called `all_threads`.
+
+By calling this function, the AMT will create a new thread instance with the given arguments, initialize the permissions, and put the function in. 
+
+After the creation of the thread, the entity will be put into `all_threads` array and wait for the utilization of the CPU.
```
DavidSpickett: So since the thread array is statically allocated it's kinda already in the array. I think you can just leave that bit out completely.

"After the creation of the thread it will wait to be chosen by the scheduler so it can utilize the CPU"?

(also good to see "the cpu" here, so the "a cpu" above should definitely be changed)

demos/yielding/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+## `do_scheduler`
+
+The job of the scheduler is to choose the next thread to make use of the CPU. 
+
+In AMT, the scheduler simply walks through the `all_threads` array to find the next thread.
+
+To run the thread chosen by the scheduler and restore the function call stack of the current thread, we need to save the information of this thread first.
```
DavidSpickett: Your description here isn't quite right. To be sure we're on the same page, here's how we get to do_scheduler most of the time: (first boot is a little different)
- user thread calls yield()
- this does syscall_yield
- we enter the kernel
- current thread context is saved
- we find that it's a syscall_yield and handle it
- by calling do_scheduler
- do_scheduler picks the next_thread
- we go back into assembly to switch the **already saved** current thread with the next one

There are times where yield.S will call do_scheduler but I don't think we need to cover them here. (I hope)

demos/yielding/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+To achieve the goal, the context switch routine written in assembly (`yield.S`) will be executed to find the next task. 
+
+And the task of the thread to be switched will be suspended until the thread has been chosen to run
+
+## `yield`
+
+To yield the utilization, the thread calls the syscall yield and set the kind of yielding as `YIELD_ANY`, which means that we do not specify the thread to acquire the CPU. As a result, the scheduler will help us to choose the next thread to run.
```
DavidSpickett: "To yield the utilization, the thread calls the syscall `yield` and sets the kind of yielding as `YIELD_ANY`. This means that we do not specify the next thread to run on the CPU. As a result, the scheduler will choose the next thread for us."

* set -> sets
* ticks around yield in syscall yield
* End sentence after YIELD_ANY
* Reword the "Which means" sentence
* Say that the scheduler will choose the next thread (which I suppose is helping, but "help" also implies the scheduler doesn't do all the work, which it does)

demos/yielding/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+And the current thread will have to wait until next time a thread had yielded the utilization to it.
```
DavidSpickett: I dislike starting sentences with "And", "The current thread" works fine.

I think "will have to wait until it is yielded to, to run again" works better. (either it's picked by the scheduler or specifically picked by thread ID, either fits that wording)

demos/yielding/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+## The progress of yielding demo
+
+At the beginning of the demo, to observe the behaviour of the scheduler, we use `set_kernel_config` to allow the scheduler to print out the log.
```
DavidSpickett: "At the beginning of the demo we use `set_kernel_config` to enable scheduler logging."

demos/yielding/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+To yield the utilization, the thread calls the syscall yield and set the kind of yielding as `YIELD_ANY`, which means that we do not specify the thread to acquire the CPU. As a result, the scheduler will help us to choose the next thread to run.
+
+And the current thread will have to wait until next time a thread had yielded the utilization to it.
+
+## The progress of yielding demo
+
+At the beginning of the demo, to observe the behaviour of the scheduler, we use `set_kernel_config` to allow the scheduler to print out the log. 
+
+After setting up the config, we use `add_thread_from_worker` to create a new thread running the `thread_worker`.
```
DavidSpickett: "Then we use `add_thread_from_worker` to create a new thread running `thread_worker`."

demos/yielding/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+And the current thread will have to wait until next time a thread had yielded the utilization to it.
+
+## The progress of yielding demo
+
+At the beginning of the demo, to observe the behaviour of the scheduler, we use `set_kernel_config` to allow the scheduler to print out the log. 
+
+After setting up the config, we use `add_thread_from_worker` to create a new thread running the `thread_worker`. 
+
+When the initialization is complete, the first thread starts to run into an endless loop and repeatedly yield to others.
```
DavidSpickett: "the first thread (`setup`)"

I confused myself reading this because I thought we had two threads in **addition** to the setup thread. Would be good to note that just for people like me who aren't paying attention :)

demos/yielding/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+Since there are only two threads, the scheduler will directly choose the thread running `thread_worker` to execute.
+
+In `thread_worker` function
```
DavidSpickett: I'm not sure of the utility of including the source in the doc directly. Do you envision these being standalone documents or being read side by side with the source of the demo?

Purely from my point of view I'd rather have them describe the source, show names and function signatures, key variables, but not chunks of code. (in general)

demos/yielding/README.md
```diff
@@ -0,0 +1,112 @@
...
+        log_event("exiting");
+        exit(0);
+      }
+      yield();
+    }
+  }
+}
+```
+
+We know that the thread will try to call `yield` twice and call `exit` to terminate the demo.
```
DavidSpickett: So if we don't include source chunks then here is a brief description of what thread_worker does and how we ping pong between thread_worker and setup as they yield. (which gets us round the for loop)

Side note: I'm not sure that even needs to have a while (1) in it. That might have been from way back when the kernel wouldn't naturally exit when all threads had finished.

(small changes like that are an argument to not include code just descriptions, IMO)

demos/yielding/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+To explain the log message of "yielding" and "resuming", we can check the implementation of `yield`.
```
DavidSpickett: Again I don't think we need the source chunk here. You could take a snippet of the log and explain where each message gets printed, from a high level.

Like "in the snippet above the message "yielding" is printed just before..." etc.

demos/yielding/README.md
```diff
@@ -0,0 +1,112 @@
...
+void yield(void) {
+  log_event("yielding");
+  DO_SYSCALL_2(yield, INVALID_THREAD, YIELD_ANY);
+  log_event("resuming");
+}
+```
+
+We can find that before yielding, the thread will dump the log message "yielding", and when the utilization is back to this thread, the log message "resuming" will be dumped on the screen and the thread will resume running the given subroutine.
+
+According to the behaviour of the two threads, the result of this demo will be:
```
DavidSpickett: Similar to the source chunks I don't think including log output is going to be useful. Assuming the reader has the repo open with the docs, they can read the expected output themselves.

I would prefer a higher level description of the demo. Here you could explain where the "yielding", "resuming" messages come from (which I think is a great idea for the first demo people will read) then give a description of the rest.

Like (super rough wording): "thread 0 repeatedly yields to thread 1. Thread 1 counts the number of times it has run. If it has run twice then it calls exit() to end the demo. This shows that threads are resumed from the point they yield from, if this were not the case then the demo would effectively become an infinite loop"

Maybe "run" is too imprecise, it can mean it was restarted (another imprecise word). So maybe "number of times it has been yielded to", which makes it clear that it doesn't restart. (or "has has had CPU time" whatever makes sense to you)

Side note: The docs for Linux PTRACE_CONT use the word "restart" which confused me a good deal, that's why I bring that up.

demos/yielding/README.md
```diff
@@ -0,0 +1,112 @@
+# Yielding
+
+In this demo, we try to show the usage of yielding, which is one characteristic of a multitasking system.
```
DavidSpickett: "In this demo we show the usage of yielding, which is one element of a multitasking system.

- remove the first comma
- we don't try we definitely do :) (whether the reader feels we succeeded is up to them)
- characteristic -> element. Characteristic seems like an adjective thing to me, like "suqishy" or "bold". Ok maybe not in this context but you get the idea.

DavidSpickett (changes requested):
This is looking good so far, thanks! I learned something new myself reading it so clearly these docs will be helpful.

I think we can get a lot of general things decided on this first document and that will save a lot of time later.

OscarShiang:
Thanks for the suggestions, I just updated the README file. 
If there is any question, please let me know : )

demos/yielding/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+The job of the scheduler is to choose the next thread to make use of the CPU. 
+
+In AMT, the scheduler simply walks through the `all_threads` array to find the next thread.
+
+Then the task of the thread to be switched will be suspended until the thread has been chosen to run.
+
+## [`yield`](https://github.com/DavidSpickett/ARMMultiTasking/blob/master/src/user/thread.c#L92)
+
+To yield the utilization, the thread calls the syscall `yield` and sets the kind of yielding as `YIELD_ANY`. The scheduler will choose the next thread to run itself.
```
DavidSpickett: "This means that the scheduler will choose the next thread to run."

- Remove "itself", kinda sounds like the scheduler runs itself
- "This means that" to connect with the mention of `YIELD_ANY`

demos/yielding/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+In this demo we show the usage of yielding, which occurs in a computer program during multithreading.
+
+By calling syscall `yield`, the thread gives away the utilization of the CPU to another thread, or a specific thread (discussed in another demo).
+
+To describe thread yielding, we need to introduce:
+
+* [`add_thread_from_worker`](https://github.com/DavidSpickett/ARMMultiTasking/blob/master/src/user/thread.c#L42)
+* [`do_scheduler`](https://github.com/DavidSpickett/ARMMultiTasking/blob/master/src/kernel/thread.c#L353)
+* [`yield`](https://github.com/DavidSpickett/ARMMultiTasking/blob/master/src/user/thread.c#L92)
```
DavidSpickett: Sorry to waste your effort but these links are cool now, until the source changes and then they point to something completely different.
The names are unique enough you could grep for them so I think it's fine to leave out the links.
(I think you can link to a file within a specific commit but still, I don't think it's worth the effort)

Maybe just put the source file name in brackets? Which is what you had before I think.

demos/yielding/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+## [`add_thread_from_worker`](https://github.com/DavidSpickett/ARMMultiTasking/blob/master/src/user/thread.c#L42)
```
DavidSpickett: Same thing for these, drop the links. (no need to mention the source file again here)

demos/yielding/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+In AMT we store all of the threads in an array called `all_threads`.
+
+When a thread calls this function, the kernel will create a new thread with the given arguments, initialize the permissions, and set its starting point to be the given function (would be `thread_worker()` in this demo).
```
DavidSpickett: "calls `add_thread_from_worker`"

Otherwise it sounds like we're calling `all_threads` somehow.

"(which is `thread_worker`"

"would be" sounds like the demo might do something else but it always uses `thread_worker`

demos/yielding/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+Then the task of the thread to be switched will be suspended until the thread has been chosen to run.
+
+## [`yield`](https://github.com/DavidSpickett/ARMMultiTasking/blob/master/src/user/thread.c#L92)
+
+To yield the utilization, the thread calls the syscall `yield` and sets the kind of yielding as `YIELD_ANY`. The scheduler will choose the next thread to run itself.
+
+The current thread will have to wait until it is yielded to, to run again.
+
+## The progress of yielding demo
```
DavidSpickett: I think there's a better way to title this section.
"Walkthrough" perhaps? We can leave out the demo's name, that's obvious from context.

demos/yielding/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+Then we use `add_thread_from_worker` to create a new thread running `thread_worker`. 
+
+When the initialization is complete, the first thread (`setup`) starts to run into an endless loop and repeatedly yields to others.
+
+Since there are only two threads, the scheduler will directly choose the thread running `thread_worker` to execute.
+
+After the thread (`thread_worker`) gets the use of the CPU, it counts the number of times it has been yielded to and yields the use. If it had been yielded to twice, it calls `exit()` to end the demo. 
+
+This shows that threads are resumed from the point they yield from, if the thread didn't go back to the point it yielded from or restore all its local variables back then the demo would effectively become an infinite loop.
```
DavidSpickett: 👍 

This format works really well.

- describe the initial state
- high level description of steps
- say what the demo is trying to prove

Two small things:
"If it ~had~ has been yielded to twice"
"or restore all its local variables ~back~ then the demo"

demos/yielding/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+To describe thread yielding, we need to introduce:
+
+* [`add_thread_from_worker`](https://github.com/DavidSpickett/ARMMultiTasking/blob/master/src/user/thread.c#L42)
+* [`do_scheduler`](https://github.com/DavidSpickett/ARMMultiTasking/blob/master/src/kernel/thread.c#L353)
+* [`yield`](https://github.com/DavidSpickett/ARMMultiTasking/blob/master/src/user/thread.c#L92)
+
+## [`add_thread_from_worker`](https://github.com/DavidSpickett/ARMMultiTasking/blob/master/src/user/thread.c#L42)
+
+In AMT we store all of the threads in an array called `all_threads`.
```
DavidSpickett: I would move/merge this line into the `do_scheduler` description. That's where it's most relevant. Here it's a bit of a non sequitur.

(roughly)
AMT stores threads in all_threads. The scheduler walks this array .. etc etc.

demos/yielding/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+After the creation of the thread, it will wait to be chosen by the scheduler so it can utilize the CPU.
+
+## [`do_scheduler`](https://github.com/DavidSpickett/ARMMultiTasking/blob/master/src/kernel/thread.c#L353)
+
+The job of the scheduler is to choose the next thread to make use of the CPU. 
+
+In AMT, the scheduler simply walks through the `all_threads` array to find the next thread.
+
+Then the task of the thread to be switched will be suspended until the thread has been chosen to run.
```
DavidSpickett: "Then the the current thread will be suspended until it is chosen to run again."

- "the current thread" is clearer than "task to be switched" for me. The latter could mean the current or the next thread.
- Remove "task of" since if the thread is suspended, then it's task/work code is implicitly also suspended
- add "again" on the end, to emphasize that it will be resuming not running for the first time.

Side note: When we make a syscall the "current thread" is technically suspended but I still call it the current thread until `current_thread` actually changes.

DavidSpickett (changes requested):

OscarShiang:
I have fixed the problems. Thanks for your hints.

DavidSpickett (approved):

DavidSpickett:
LGTM, thanks!

PR #7, opened 2020-11-15
Create README.md in demos/exyielding

OscarShiang:
Explain the mechanism in exyielding demo.

### DIFF
--- a/demos/exyielding/CMakeLists.txt
+++ b/demos/exyielding/CMakeLists.txt
@@ -1,1 +1,1 @@
-add_demo(exyielding 3)
+add_demo(exyielding 4)
--- a/demos/exyielding/README.md
+++ b/demos/exyielding/README.md
@@ -1,0 +1,27 @@
+# Exyielding
+
+In this demo we want to show that a thread can yield the utilization of the CPU directly to another thread.
+
+To illustrate the mechanism in it, we will explain:
+
+* `yield_to` (in `src/user/thread.c`)
+
+## `yield_to`
+
+When this function is called, the kernel will check if the thread we specified is available to yield to.
+
+If the thread can be yielded to, we will set `next_thread` to the address of this thread and call the `do_scheduler` function.
+
+Since we have decided `next_thread`, the scheduler will do housekeeping tasks and help us to setup the thread we had already chosen.
+
+If the thread we specified is unavailable to yield, `do_scheduler` will not be called and we will go back to the point we called `yield_to()`.
+
+## Walkthrough
+
+At the beginning of this demo, we create 3 named threads `first`, `second` and `last`. After the thread `setup` ends, the thread `first` will get the use of the CPU and yield to the thread `second` specified by its thread ID (`tid`).
+
+When the thread `second` gets the use, it yields to `first` and `last` respectively and the thread `first` exits normally. This shows that we can yield to a lower `tid` as well as a higher one.
+
+After the thread `second` does `yield()`, the thread `last` will reach its end. At this point, there is only thread `second` in the `all_threads` array. So the next time it tries to do `yield_to()` or `yield()`, it should fail and thus yielding will not occur.
+
+As a result, the thread `second` will exit normally, and the kernel will also exit since there are no more threads to run.
--- a/demos/exyielding/expected.log
+++ b/demos/exyielding/expected.log
@@ -9,8 +9,6 @@ Thread  <scheduler>: scheduling new thread
 Thread  <scheduler>: next thread chosen
 Thread       second: resuming
 Thread       second: yielding
-Thread  <scheduler>: scheduling new thread
-Thread  <scheduler>: next thread chosen
 Thread         last: yielding
 Thread       second: resuming
 Thread       second: yielding
--- a/demos/exyielding/exyielding.c
+++ b/demos/exyielding/exyielding.c
@@ -1,15 +1,14 @@
 #include "common/assert.h"
 #include "user/thread.h"
 
-void thread1() {
+void first() {
   yield_to(2);
 }
 
-void thread2() {
+void second() {
   yield_to(1);
-  add_named_thread(thread1, "last");
-  yield(); // switch to "last"
-  yield(); // switch back to "last"
+  yield_to(3); // switch to "last"
+  yield();     // switch back to "last"
   // Run ourselves again, shouldn't actually do a switch
   assert(!yield());
   assert(!yield_to(1));
@@ -18,6 +17,7 @@ void thread2() {
 void setup(void) {
   set_kernel_config(KCFG_LOG_SCHEDULER, 0);
 
-  add_named_thread(thread1, "first");
-  add_named_thread(thread2, "second");
+  add_named_thread(first, "first");
+  add_named_thread(second, "second");
+  add_named_thread(first, "last");
 }


demos/exyielding/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+In this demo we want to show that a thread can yield the utilization of the CPU directly to another thread.
+
+To illustrate the mechanism in it, we will explain the following:
+* `yield_to` (in `src/user/thread.c`)
+* `yield_next` (in `src/user/thread.c`)
+
+## `yield_to`
+
+By calling this function, the kernel will check if the thread we specified is available to yield to.
```
DavidSpickett: "When this function is called, ..."

demos/exyielding/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+If the thread can be yielded, we will set `next_thread` as the address of this thread and trigger `do_scheduler` function.
```
DavidSpickett: "can be yielded to"
"to the address of"
"and call the `do_scheduler` function."

demos/exyielding/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+Since we have decided `next_thread`, the scheduler will do nothing and the kernel will switch the thread.
```
DavidSpickett: Worth noting that the scheduler is also the point where we do housekeeping tasks that must be done per switch.
Like swapping code pages. So it does nothing apart from what's needed to setup the thread we already chose.

(I might refactor this part of the code to make this more obvious. Pretty sure that there's a bug if you send a signal to a thread that is then yielded directly to.)

demos/exyielding/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+If the thread we specified is unavailable to yield, `do_scheduler` will not be triggered and we will go back to the point we called `yield_to()`.
+
+## `yield_next`
+
+This function will yield the utilization of the CPU to the next valid thread. The behaviour of this function seems to be similar to the one of `yield()`, but they are a little bit different.
```
DavidSpickett: I went to verify this and was confused too.

I think yield_next is actually the same behavior as yield. The scheduler will also return to the current thread if it doesn't find any new thread, but I think there was a point where it did not.

The only difference is that yield_next tells you whether you did yield or not. Where yield just returns void. So for now I'd stick with that as the difference. It allows you to check whether there is another valid thread. (the utility of that is up for debate but that's what it is)

At some point yield should be changed to return a bool and yield_next removed.

So yeah, yield_next is just yield but it tells you whether you actually yielded.

demos/exyielding/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+When we calling `yield()`, the scheduler will be triggered and choose the next thread to run. The scheduler may choose the same thread to yield. After switching the context, the thread will go back to the point it yielded before.
+
+In case we are using `yield_next()`, the kernel will walk through `all_threads` array to find the next valid thread to yield. If there is no suitable thread, we will go back to the point we called `yield_next()`. This means that we will not enter `do_scheduler` function and that the context switch will not occur.
```
DavidSpickett: To continue the above, they both go through k_do_yield which calls do_scheduler so the housekeeping still happens.

The context switch also still happens (well, kinda) since to make a syscall the current thread context is completely saved anyway. So in a sense when we return we "switch" back in that context.

Previously I had some minimal register set being saved for syscalls but it got to be a pain making special cases for it.

demos/exyielding/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Walkthrough
+
+At the beginning of this demo, we create two named thread  (`first` and `second`). After the thread (`setup`) ends, the thread (`first`) will get the use of the CPU and yield to the thread `second` specified by its `tid`.
```
DavidSpickett: Usually for acronyms I'd say:
"specified by it's thread ID (`tid`)"

Then later you can just use the short version. (and yes I see the irony of ID also being an acronym but I think we can assume people know that one :) )

demos/exyielding/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+The thread (`second`) will use `yield_to()` and `yield_next()` to specify the thread to yield. After yielding twice to both threads (`first` and `last`), the two would have reached their end. So the next time the thread `second` tries to do `yield_to()` or `yield_next()`, it should fail and the switch will not occur. Otherwise, it will be an assertion error.
```
DavidSpickett: Remove "Otherwise...". We don't want to be listing every assert.

However, I would add something like:
"So the thread `second` will exit normally, then the kernel will exit because there are no more threads to run."
To complete the explanation. (and emphasise that not yielding is not an error, it's expected)

DavidSpickett (changes requested):
As I said I'm tempted to remove yield_next but I won't do that right now.

Let's get this reviewed and in, then we can just update the README if it is removed.

DavidSpickett:
Ok so I said I wasn't gonna remove yield_next but I had time on my hands so I did.

yield() now replaces yield_next. So the difference between here and the yielding demo is that in exyielding we do check the return value.

OscarShiang:
The changes had been updated, and I removed the description of `yield_next()`.

OscarShiang:
The changes have been updated, and I also have removed the description of `yield_next()`.

demos/exyielding/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+If the thread can be yielded to, we will set `next_thread` to the address of this thread and call the `do_scheduler` function.
+
+Since we have decided `next_thread`, the scheduler will do housekeeping tasks and help us to setup the thread we had already chosen.
+
+If the thread we specified is unavailable to yield, `do_scheduler` will not be called and we will go back to the point we called `yield_to()`.
+
+## Walkthrough
+
+At the beginning of this demo, we create two named thread  (`first` and `second`). After the thread (`setup`) ends, the thread (`first`) will get the use of the CPU and yield to the thread `second` specified by its thread ID (`tid`).
```
DavidSpickett: Remove the double space in "named thread  (".

Also you can remove the brackets around setup/first/second. Reads fine without them.

demos/exyielding/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+The thread `second` will use `yield_to()` to specify the thread to yield and use `yield()` to yield the use to other threads. After yielding twice to both threads (`first` and `last`), the two would have reached their end. The next time the thread `second` tries to do `yield_to()` or `yield()`, it should fail so the yielding will not occur.
+
+As a result, the thread `second` will exit normally, and the kernel will exit since there are no more threads to run.
```
DavidSpickett: This is missing where `last` is created.

Granted, I can't remember why I added `last` to the demo. I think perhaps to check you could yield (aka yield next) to a thread with a higher tid as well as a lower tid. Which proves that we can walk `all_threads` around to the end and back to an earlier thread, or stop at a later thread. (I don't think the explanation needs to include this detail though)

DavidSpickett (changes requested):

OscarShiang:
Ok, I have updated the changes again.

demos/exyielding/README.md
OscarShiang: I am thinking that maybe we can create the thread `last` first then do yielding to the thread `first`. It will be more clear to show that we can yield to the higher or the lower `tid` as well.  Would it be better?

demos/exyielding/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+If the thread can be yielded to, we will set `next_thread` to the address of this thread and call the `do_scheduler` function.
+
+Since we have decided `next_thread`, the scheduler will do housekeeping tasks and help us to setup the thread we had already chosen.
+
+If the thread we specified is unavailable to yield, `do_scheduler` will not be called and we will go back to the point we called `yield_to()`.
+
+## Walkthrough
+
+At the beginning of this demo, we create two named thread `first` and `second`. After the thread `setup` ends, the thread `first` will get the use of the CPU and yield to the thread `second` specified by its thread ID (`tid`).
```
DavidSpickett: thread -> threads

demos/exyielding/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+The thread `second` will first yield to the thread `first`, then the thread `first` will end. When thread `second` is yielded to again, it creates a new thread `last`. This thread is used to show that we can yield to higher `tid`  as well as the lower one.
```
DavidSpickett: (just gonna quote you here for convenience)

> I am thinking that maybe we can create the thread last first then do yielding to the thread first. It will be more clear to show that we can yield to the higher or the lower tid as well. Would it be better?

I remembered the original reason the place we add the "last" thread.
It was so that we could `yield_next` and go from:
```
[setup (exited)][first][second (current)][(invalid)]
```
to:
```
[setup (exited)][first (current)][second][(invalid)]
```
Without having `last` on the end there to get in the way.

Anyway.

> Would it be better?

Absolutely. I would add all the threads in setup() and also rename the first/second worker functions to first() and second() just to be a bit clearer. Make one of the yields here:
```
  yield(); // switch to "last"
  yield(); // switch back to "last"
```
A yield _to.

DavidSpickett: In general this demo has gone from showing `yield_next` and `yield_to` to just `yield_to` so any change to focus on the latter is fine.

DavidSpickett (changes requested):

demos/exyielding/README.md
OscarShiang: Shall we work on changing this demo first?

OscarShiang:
I have updated and fixed the typo.

demos/exyielding/README.md
DavidSpickett: Sure, do it as the first commit of this PR.

OscarShiang:
The demo has been modified with the following changes:
  - modify the names of the functions. ("thread1" to "first" and "thread2" to "second")
  - change one of the `yield()` to `yield_to`.
  - adjust the `MAX_THREADS` from 3 to 4.

The README.md has also been updated in the second commit.

demos/exyielding/CMakeLists.txt
```diff
@@ -1 +1 @@
-add_demo(exyielding 3)
+add_demo(exyielding 4)
```
DavidSpickett, line 1: I probably should have mentioned this, hope it wasn't too hard to find.

demos/exyielding/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+If the thread we specified is unavailable to yield, `do_scheduler` will not be called and we will go back to the point we called `yield_to()`.
+
+## Walkthrough
+
+At the beginning of this demo, we create 3 named threads `first`, `second` and `last`. After the thread `setup` ends, the thread `first` will get the use of the CPU and yield to the thread `second` specified by its thread ID (`tid`).
+
+When the thread `second` gets the use, it yields to `first` and `last` respectively and the thread `first` exits normally. This shows that we can yield to a lower `tid` as well as a higher one.
+
+After the thread `second` do `yield()`, the thread `last` will reach its end. At this point, there is only thread `second` in `all_threads` array, so the next time it tries to do `yield_to()` or `yield()`, it should fail and thus yielding would not occur.
```
DavidSpickett: "does `yield()`"

"At this point there is only thread `second` in the `all_threads` array. So the next time it tries to do `yield_to()` or `yield()`, it should fail and thus yielding will not occur."

in `all_threads` -> in the `all_threads`
Split into two sentences
would not occur -> will not occur

DavidSpickett (changes requested):
Minor comments, otherwise looks good. The demo is more readable with `last` moved, which I like.

OscarShiang:
OK, I have just revised the README.md.

DavidSpickett:
LGTM

PR #9, opened 2020-11-22
Add README.md into demos/message

OscarShiang:
Illustrate how this demo works and the functions' usage.

### DIFF
--- a/demos/message/README.md
+++ b/demos/message/README.md
@@ -1,0 +1,32 @@
+# Message
+
+In this demo we will show the message passing mechanism. Since messages are not sharable to other threads, this is a way for a thread to communicate with others privately.
+
+To explain this demo, we will first introduce the following:
+
+* `send_msg` (in `src/user/thread.c`)
+* `get_msg` (in `src/user/thread.c`)
+
+## `send_msg`
+
+In AMT, each thread owns a ring buffer to store pending messages. When this function is called, the kernel will check the corresponding buffer of the destination thread.
+
+If the buffer is not full, the message will be copied into the destination thread's buffer according to the `destination` and `message` we specified and remains there until the thread takes it out.
+
+If the buffer is full, the kernel will refuse to send the message, and thus the return value of this function will be `false`.
+
+## `get_msg`
+
+When a thread calls this function, the kernel checks whether there is a message in the buffer.
+
+If the buffer is non-empty, the kernel will copy the information of the message to the variables we gave and remove it from the message buffer. So we can view the `content` and `src` variables to know the content and which thread sent this message.
+
+## Walkthrough
+
+First, the `setup` thread creates two threads `sender` and `receiver` then it jumps to the `spammer` function.
+
+`spammer` doesn't end or give away the use of the CPU, so it repeatedly sends messages to `receiver`, which makes `receiver`'s message buffer full. As a result, when `sender` runs, it fails to send a message to `receiver`.
+
+When `receiver` runs, it drops the messages which are not from `sender`. Since the message buffer only contains the messages from `spammer`, they will all be dropped and `receiver` will yield.
+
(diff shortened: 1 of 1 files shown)


demos/message/README.md
```diff
@@ -0,0 +1,32 @@
...
+To explain this demo, we will first introduce the following:
+
+* `send_msg` (in `src/user/thread.c`)
+* `get_msg` (in `src/user/thread.c`)
+
+## `send_msg`
+
+In AMT, each thread owns a ring buffer to store pending messages. When this function is called, the kernel will check the corresponding buffer of the destination thread.
+
+If the buffer is not full, the message will be copied into the destination thread's buffer according to the `destination` and `message` we specified and wait for the thread to take it out.
```
DavidSpickett: "we specified and remains there until the thread takes it out."

demos/message/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+If the buffer is full, the kernel will refuse to send the message, and thus the return value of this function will be `false`.
+
+## `get_msg`
+
+When a thread calls this function, the kernel checks whether there is a message in the buffer.
+
+If the buffer is non-empty, the kernel will copy the information of the message to the variables we gave. So we can view the `content` and `src` variables to know the content and which thread sent this message.
```
DavidSpickett: This is good but note that it also frees a slot in the messages:
"to the variables we gave and remove it from the message buffer."

demos/message/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+When a thread calls this function, the kernel checks whether there is a message in the buffer.
+
+If the buffer is non-empty, the kernel will copy the information of the message to the variables we gave. So we can view the `content` and `src` variables to know the content and which thread sent this message. 
+
+## Walkthrough
+
+First, the `setup` thread creates two threads `sender` and `receiver` then it jumps to the `spammer` function.
+
+Because it doesn't end or give away the use of the CPU, it repeatedly sends messages to `receiver`, which makes `receiver`'s message buffer full. As a result, `sender` fails to send any message to `receiver` when it gets the use of the CPU.
```
DavidSpickett: Starting with because is generally not good (though I do it myself all the time I must admit). I think this would sound better:
"`spammer` doesn't end or give away use of the CPU, so it repeatedly"

"As a result, `sender` fails to send any message to `receiver` when it gets the use of the CPU."

Would read better as:
"As a result, when `sender` runs it fails to send a message to `receiver`."

Unless it makes things unclear (which I don't think it does here) using "runs" instead of "gets use of the CPU" is quicker and simpler.

demos/message/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+When the CPU is allocated to `receiver`, `receiver` drops the messages which are not from `sender`. Since the message buffer only contains the messages from `spammer`. The `receiver` thread will drop all of them and call `yield()`.
```
DavidSpickett: Same here, "when `receiver` runs" would be simpler.

You can combine the last sentence:
"Since the message buffer only contains the messages from `spammer`they will all be dropped and `receiver` will yield."

Also you can use yield as a plain verb too I think. Saves some typing.

demos/message/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+After `spammer` has clogged the `receiver`'s buffer twice, it sends only one message and reaches its end. At this time, `sender` can send the message to `receiver` as the buffer is not full. So `receiver` will get the message sent from `sender` and exit this demo normally.
```
DavidSpickett: It would be better to link this back to the previous text by saying:
"`spammer` will clog `receiver`'s buffer again and the process repeats. Finally `spammer` will send only a single message, then exit."

"So `receiver` will get the message sent from `sender` and end the demo."

Seems weird to say "exit normally" since we've not referenced an exit that isn't normal.

DavidSpickett (changes requested):

OscarShiang:
OK, I just revised the file.

DavidSpickett:
LGTM

PR #11, opened 2020-11-26
Remove k_close in k_stdout_isatty

### LINKED ISSUE #10: Fail to run demos while using QEMU 4.2.1
**Describe the bug**
The shell demo freezes at the beginning and fails to run. Other demos finish without any output. 

**Configuration:**
 - cmake version (`cmake --version`)
```
cmake version 3.16.3
```
 - toolchain/compiler version (`gcc --version`, for example)
```
arm-none-eabi-gcc (GNU Arm Embedded Toolchain 9-2020-q2-update) 9.3.1 20200408 (release)

aarch64-none-elf-gcc (GNU Toolchain for the A-profile Architecture 9.2-2019.12 (arm-9.10)) 9.2.1 20191025
```
 - QEMU version (`qemu-system-arm --version`)
```
QEMU emulator version 4.2.1 (Debian 1:4.2-3ubuntu6.8)
```

**To Reproduce:**
- cmake configuration
```
-- BUILD_PLATFORM is arm
-- OPT_LEVEL is -O0
-- SANITIZERS are ON
-- LTO is OFF
-- COVERAGE is 
-- STACK_SIZE is 2KB
-- ccache is enabled
-- Configuring done
-- Generating done
-- Build files have been written to: /home/oscar/ARMMultiTasking
```
- make command
```
make
```
- run command
```
make run_shell
```

**Expected behavior**
The demos will print out some log messages or let us to type something. 

**Affected platforms/settings**
It occurs on all platforms:
- arm
- thumb
- aarch64

**Additional context**
I think that it is a problem when using QEMU 4.2.1, since it successfully runs on QEMU 5.1.0. 

I use `git checkout` to some previous commit. It seems that the problem occurs after `b089ef2`. 

I try to use gdb to locate the error, but it turns out the shell demo works properly when I execute the code one by one. 
So I am not sure where the error is.

OscarShiang:
Fix #10.

It is a workaround to avoid the outputting error in QEMU 4.2.1.

Since the stdout of QEMU may be closed when we close the stdout in
our guest platform, we leave the stdout opened to make sure that the
stdin/stdout are still available to use.

This commit has been tested in QEMU 5.1.0 and QEMU 4.2.1 and both of them run properly.

### DIFF
--- a/src/kernel/file.c
+++ b/src/kernel/file.c
@@ -50,7 +50,8 @@ bool k_stdout_isatty(void) {
   }
 
   int istty = k_isatty(stdout_fd);
-  k_close(stdout_fd);
+  // Closing stdout_fd may cause QEMU 4.2.1 to close its stdout
+  // So we just leave it opened here
 
   return istty == 1;
 }


jserv:
Typo. Should be "workaround" instead of "walkaround."

OscarShiang:
Sorry about that.  I have force-pushed it again.

DavidSpickett:
Walkaround does sound more relaxing though :)

LGTM thanks for fixing this.

PR #12, opened 2020-12-02
Add README.md to describe the exit demo

OscarShiang:
Create some description to show the mechanism of the exit demo.

In this demo, I use `setup` thread to indicate thread 0 at the beginning. 
After it calls the `counter()` function, I use `counter` to describe the thread instead.

### DIFF
--- a/demos/exit/README.md
+++ b/demos/exit/README.md
@@ -1,0 +1,49 @@
+# Exit
+
+In this demo we are going to show that threads can exit normally like C functions. We can use functions like `thread_join` to wait for specific threads to exit.
+
+The exit we will illustrate here is not the function call `exit()`. This function causes the termination of the whole system while the one we want to show is the behaviour after a thread finishes its worker routine.
+
+To explain the demo, we need to introduce:
+* `make_args` (in `include/common/thread.h`)
+* `add_named_thread_wtih_args` (in `src/user/thread.c`)
+* `thread_start` (in `src/kernel/thread.c`)
+* `thread_join` (in `src/user/thread.c`)
+
+## `make_args`
+
+In AMT, we use `ThreadArgs` to pack arguments so we can pass these variables into the thread we want to create. 
+
+Since `ThreadArgs` comes with 4 variables, we can pass at most 4 arguments into a thread. On the other hand, if we want to use less than 4 arguments, we need to provide 0 as dummy argument.
+
+`make_args` helps us to transform the assignment into a form like a function call and make it more readable.
+
+## `add_named_thread_with_args`
+
+`add_named_thread_with_args` creates a thread with the given `name`, `args` and `worker` routine.
+
+The behaviour is the same as `add_thread_from_worker()`, except that we can pass arguments to the thread created. (`add_thread_from_worker` sets all arguments to 0)
+
+## `thread_start`
+
+It is the starting point of every thread since the initial PC has been set to the address of this function.
+
+After entering this function, the thread begins to run the worker routine we defined before. 
+
+When the routine ends successfully, `thread_start` sets the state of the thread to `finished` then switches to other threads.
+
+## `thread_join`
+
+The `thread_join()` function waits for the thread specified by `tid` to terminate. When a thread calls this function, it keeps yielding until the target thread exits. 
+
+If the value of `state` is not `NULL`, then `thread_join()` copies the exit state of the target thread into the location pointed to by `state`.
+
+## Walkthrough
+
+For starters, we create 2 threads and pass `2` and `4` into the corresponding threads. This is the number of times each thread will yield before exiting. 
+
+After the creation, the `setup` thread jumps into the `counter()` function and changes its name to `counter`. It waits for all threads apart itself to exit. Since all of the threads do not exit, the `counter` thread will repeatedly yield.
+
+Thread 1 will reach its end after it yields twice. So the `counter` thread should receive the state of thread 1 as `finished` and start to wait for thread 2. At this time, thread 2 has two yields left before it will exit.
+
+When thread 2 exits, `thread_join` will set `state` to `finished`. Then the scheduler chooses `counter` thread to yield to. Since both of the threads exited, the `counter` will also exit and end the demo.
--- a/demos/exit/exit.c
+++ b/demos/exit/exit.c
@@ -9,6 +9,7 @@ void work(int num) {
 }
 
 void counter() {
+  set_thread_name(CURRENT_THREAD, "counter");
   // Start at 1 so we don't join on ourselves
   for (int i = 1; i < 3; ++i) {
     ThreadState state = init;
@@ -25,6 +26,5 @@ void setup(void) {
   ThreadArgs ta2 = make_args(4, 0, 0, 0);
   add_named_thread_with_args(work, NULL, &ta2);
 
-  yield();
   counter();
 }
--- a/demos/exit/expected.log
+++ b/demos/exit/expected.log
@@ -1,27 +1,27 @@
-Thread            0: yielding
+Thread      counter: yielding
 Thread            1: yielding
 Thread            2: yielding
-Thread            0: resuming
-Thread            0: yielding
+Thread      counter: resuming
+Thread      counter: yielding
 Thread            1: resuming
 Thread            1: yielding
 Thread            2: resuming
 Thread            2: yielding
-Thread            0: resuming
-Thread            0: yielding
+Thread      counter: resuming
+Thread      counter: yielding
 Thread            1: resuming
 Thread            1: exiting
 Thread            2: resuming
 Thread            2: yielding
-Thread            0: resuming
-Thread            0: thread 1 exited
-Thread            0: yielding
+Thread      counter: resuming
+Thread      counter: thread 1 exited
+Thread      counter: yielding
 Thread            2: resuming
 Thread            2: yielding
-Thread            0: resuming
-Thread            0: yielding
+Thread      counter: resuming
+Thread      counter: yielding
 Thread            2: resuming
 Thread            2: exiting
-Thread            0: resuming
-Thread            0: thread 2 exited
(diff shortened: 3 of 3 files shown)


demos/exit/README.md
```diff
@@ -0,0 +1,38 @@
+# Exit
+
+In this demo we are going to show that threads can exit normally like C functions. So we can use functions like `thread_join` to let a thread to wait for some threads to exit.
```
DavidSpickett: This:
"So we can use functions like `thread_join` to let a thread to wait for another thread to exit. "

Would be clearer as:
"We can use functions like `thread_join` to wait for specific threads to exit."

demos/exit/README.md
```diff
@@ -0,0 +1,38 @@
...
+To explain the demo, we need to introduce:
+* `make_args` (in `include/common/thread.h`)
+* `add_named_thread_wtih_args` (in `src/user/thread.c`)
+* `thread_join` (in `src/user/thread.c`)
+
+## `make_args`
+
+In AMT, we use `ThreadArgs` to pack arguments so we can pass these variables into the thread we want to create. 
+
+Since `ThreadArgs` comes with 4 variables, we need to create the variable before we pass it into the thread.
```
DavidSpickett: I'm not sure what this sentence is getting at. Perhaps you are saying that a thread can have up to 4 arguments, so if you're using <4 you provide 0 as a dummy argument?

Or do you mean to say that you must make the ThreadArgs first, then make the thread?

(both seem worth mentioning)

demos/exit/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+`make_args` helps us to transform the assignment into a form like a function call and make the assignment of the `ThreadArgs` becomes more readable.
```
DavidSpickett: "and make it more readable."

It's implied that you're referring to "the assignment" from earlier in the sentence.

demos/exit/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+## `add_named_thread_with_args`
+
+`add_named_thread_with_args` creates a thread with the given `name`, `args` and `worker` routine.
+
+The behaviour is just like what `add_thread_from_worker()` does, but the arguments of the threads created by `add_thread_from_worker()` are initialized to 0 while we can pass the arguments into the threads created by `add_named_thread_with_args`.
```
DavidSpickett: This is a rather long sentence. How about:
"The behaviour is the same as `add_thread_from_worker()`, except that we can pass arguments to the thread created. (`add_thread_from_worker()` sets all arguments to 0)"

Side note: Firefox is complaining to me that "behaviour" is spelt wrong because of the "ou" instead of the American "behavior".  (because I'm lazy and didn't change my dictionary)

I use "ou" myself but I don't mind either way you want to spell that class of words. (I hadn't noticed it until now)

demos/exit/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+## `thread_join`
+
+The `thread_join()` function waits for the thread specified by `tid` to terminate. When a thread calls this function, it keeps yielding until the target thread exits. 
+
+If the value of `state` is not NULL, then `thread_join()` copies the exit state of the target thread into the location pointed to by `state`.
```
DavidSpickett: I would quote/format NULL like `NULL`

demos/exit/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Walkthrough
+
+For starters, we create 2 threads and pass `2` and `4` into the corresponding threads. After the creation, the `setup` thread jumps into the `counter()` function.
+
+In the `counter()` function, it uses `thread_join` to wait thread 1. Since the thread does not exit, the `setup` thread will repeatedly yield.
```
DavidSpickett: "to wait on thread 1"

demos/exit/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+If the value of `state` is not NULL, then `thread_join()` copies the exit state of the target thread into the location pointed to by `state`.
+
+## Walkthrough
+
+For starters, we create 2 threads and pass `2` and `4` into the corresponding threads. After the creation, the `setup` thread jumps into the `counter()` function.
+
+In the `counter()` function, it uses `thread_join` to wait thread 1. Since the thread does not exit, the `setup` thread will repeatedly yield. 
+
+Thread 1 will reach its end after it yields twice. So the `counter` thread should receive the state of thread 1 as `finished`, and start to wait for thread 2. At this time, thread 2 has two more times to yield.
```
DavidSpickett: You can remove the comma:
"So the `counter` thread should receive the state of thread 1 as `finished` and start to wait for thread 2."

"At this time, thread 2 has two yields left before it will exit."

demos/exit/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+When the thread 2 exits, `thread_join` should set the `state` to `finished`. Then the scheduler chooses `counter` thread to yield to. Since both of the threads exited, the `counter` will also exit and end the demo.
```
DavidSpickett: You can remove the "the", also should -> will:
"will set `state` to `finished`."

demos/exit/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+## `thread_join`
+
+The `thread_join()` function waits for the thread specified by `tid` to terminate. When a thread calls this function, it keeps yielding until the target thread exits. 
+
+If the value of `state` is not NULL, then `thread_join()` copies the exit state of the target thread into the location pointed to by `state`.
+
+## Walkthrough
+
+For starters, we create 2 threads and pass `2` and `4` into the corresponding threads. After the creation, the `setup` thread jumps into the `counter()` function.
```
DavidSpickett: I think it would be worth saying here what the numbers are for and what counter does overall. Like:
"For starters, we create 2 threads and pass `2` and `4` into the corresponding threads.
This is the number of times each thread will yield before exiting."

"After the creation, the `setup` thread jumps into the `counter()` function. This waits for all threads apart from itself to exit."

(then modify the sentence below to account for that)

demos/exit/README.md
```diff
@@ -0,0 +1,38 @@
+# Exit
+
+In this demo we are going to show that threads can exit normally like C functions. So we can use functions like `thread_join` to let a thread to wait for some threads to exit. 
+
```
DavidSpickett, line 6: I think you should note here the difference between exiting the kernel and a thread exiting. (or finishing...the terminology overlap gets a bit weird)

Something like:
"This is in contrast to exiting using `exit()` which exits the whole system."

Also I think `thread_start` should be one of the introduced functions. Noting that:
* threads start here before they run their worker function
* by doing so we catch it's return and set its state to finished and go back to the kernel

You don't need to mention it in the walkthrough but I think it's good to include a bit of the detail of how we handle threads returning.

DavidSpickett (changes requested):
I had an idea about these docs. It would be cool to add a page to docs/ that links to all the ones so far, to give a rough reading order. (the same order you're adding them)

like:
* a quick bit about how you should read the demo's readme side by side with the code
* links to each one in some "reading order"

If you made this file docs/demos.md that would be great. At some point I do want to document how the tests work and I could link to that description from there.
And of course we can link to that file from the main readme too.

You don't have to include that in this change, just something that occurred to me.

demos/exit/README.md
OscarShiang: > Also I think thread_start should be one of the introduced functions.

Thanks for pointing out this. I didn't notice that before, and I think it is really helpful to understand the change of the state in this demo.

OscarShiang: > Also I think thread_start should be one of the introduced functions. 

Thanks for pointing out this. I didn't even notice that before...

OscarShiang:
I think we can set the name of the initial thread (thread 0) to `counter` to seperate thread 0 and the other 2 threads. 

Shall we work on this change?

OscarShiang:
> It would be cool to add a page to docs/ that links to all the ones so far, to give a rough reading order. (the same order you're adding them)

Sounds great. I think I will add it with another PR.

OscarShiang:
I updated the commit. This commit includes the introduction of `thread_start`.

demos/exit/README.md
```diff
@@ -0,0 +1,49 @@
...
+The `thread_join()` function waits for the thread specified by `tid` to terminate. When a thread calls this function, it keeps yielding until the target thread exits. 
+
+If the value of `state` is not `NULL`, then `thread_join()` copies the exit state of the target thread into the location pointed to by `state`.
+
+## Walkthrough
+
+For starters, we create 2 threads and pass `2` and `4` into the corresponding threads. This is the number of times each thread will yield before exiting. 
+
+After the creation, the `setup` thread jumps into the `counter()` function. It waits for all threads apart itself to exit. Since all of the threads do not exit, the `counter` thread will repeatedly yield.
```
DavidSpickett: So setup actually yields once before going to `counter()` but I think we should remove that call, it doesn't add anything.

DavidSpickett:
> I think we can set the name of the initial thread (thread 0) to `counter` to seperate thread 0 and the other 2 threads.
> 
> Shall we work on this change?

Yeah do that and also remove the `yield()` call in setup, it's not adding anything.

Docs look good for the current state, apart from the `yield()` detail.

OscarShiang:
OK, I have modified the demo and split it into 2 commits.

DavidSpickett (changes requested):
LGTM

DavidSpickett:
Took me a bit to wrap my head around why removing the yield() didn't change the demo output more but I eventually got there. Another sign it was a good all to remove it!

DavidSpickett (approved):

DavidSpickett:
Btw, I've been thinking that having separate tests as well as the demos might be a good move. There's some things that are just randomly jammed into demos where they don't really fit.

If you see stuff like that please call it out.

PR #13, opened 2020-12-05
Add documentation of demos

OscarShiang:
Add a document into docs/

Although I am still working on the README files of the demos, I think I will just put the links here.
I hope that we can save some time so that we don't have to revise this file whenever we add new README files :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,22 @@ Current targets are:
 
 For implementation details see the [design doc](docs/design.md).
 
+## Features
+
+* Cooperative multitasking
+* Timer based thread switching
+* Loading ELF programs, including position independent code
+* Message passing
+* Mutexes and condition variables
+* Thread local storage
+* Memory allocation
+* File handling (via semihosting)
+* Signal handling
+* Permissions and errno for error handling
+* Access to thread registers for tracing
+* User space threads (fibres)
+* Stack under/overflow detection
+
 ## Building
 
 Install cmake, make and an arm-none-eabi or aarch64-none-elf toolchain. Arm developer releases are the best way to get these.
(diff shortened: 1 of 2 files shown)


OscarShiang:
By the way, should we put some brief description after each demo link?

docs/demos.md
```diff
@@ -0,0 +1,29 @@
+# AMT Demos 
+
+These documentations give an overview of how the demos work and explain some mechanism in these demos. (including some functions we use in the demo)
```
DavidSpickett: "The documents listed below give"
"explain some mechanisms in these"

You can remove "(including some functions we use in the demo)" I think it's implied already.

docs/demos.md
```diff
@@ -0,0 +1,29 @@
...
+
+For a better understanding, we suggest that you can try to revise some code to change and verify the behaviour of the demos or dig into them by using commands like `make debug` and trace them one by one.
```
DavidSpickett: I like this but the format is a bit awkward. I'd do something like this to be clearer:

For the best understanding we suggest that you:
* read the document and demo source side by side
* make some changes to the demo to verify its behaviour
* follow it in a debugger with `make debug_<demo>`

Btw I hadn't thought of the last two points, good ideas.

You can link to the debugging doc there too https://github.com/DavidSpickett/ARMMultiTasking/blob/master/docs/debugging.md

docs/demos.md
```diff
@@ -0,0 +1,29 @@
...
+* [alloc](../demos/alloc/README.md)
+* [loadbinary](../demos/loadbinary/README.md)
+* [loadbinaries](../demos/loadbinaries/README.md)
+* [loadpiebinary](../demos/loadpiebinary/README.md)
+* [parentchild](../demos/parentchild/README.md)
+* [permissions](../demos/permissions/README.md)
+* [trace](../demos/trace/README.md)
+* [signalhandling](../demos/signalhandling/README.md)
+* [backtrace](../demos/backtrace/README.md)
+* [fibres](../demos/fibres/README.md)
```
DavidSpickett: I'd rather you list just the ones that have been done. (assume that I'll merge this after spawn goes in)

DavidSpickett (changes requested):
> By the way, should we put some brief description after each demo link?

There is a table in the main readme. I have been thinking about removing that. So what you could do is move that table into this doc and make the first column names link to the documents. (then all you have to do for a new doc is to add the link, the name/descriptions are already done for you)

Then add a "Features" section before "Building" in the main readme. That lists the top level features. I think this list would be:
* Cooperative multitasking
* Timer based thread switching
* Loading ELF programs, including position independent code
* Message passing
* Mutexes and condition variables
* Thread local storage
* Memory allocation
* File handling (via semihosting)
* Signal handling
* Permissions and errno for error handling
* Access to thread registers for tracing
* User space threads (fibres)
* Stack under/overflow detection

And anything you think I've missed. Keeping it to top level things. E.g. the ability to cancel a thread makes a good demo but I wouldn't list it as a top level "feature" since it's kinda implied.

docs/demos.md
DavidSpickett: (but if you move the table here you can just link the first few names)

OscarShiang:
All right, I have adjusted them.

OscarShiang:
OK, I have adjusted them.

docs/demos.md
```diff
@@ -0,0 +1,34 @@
+# AMT Demos
+
+The documents listed below give an overview of how the demos work and explain some mechanisms in these demos.
```
DavidSpickett: Reading it again, I think "linked" makes more sense than "listed".

README.md
```diff
@@ -53,34 +69,6 @@ Each demo has `run_<demo>`, `debug_<demo>` and `test_<demo>` make targets.
 
 To test all the demos use `lit` (https://pypi.org/project/lit/) or just run `make`.
```
DavidSpickett, line 71: "For a full list see <link to demo doc>."

DavidSpickett (changes requested):

OscarShiang:
Alright, I updated it again.

DavidSpickett (approved):
Thanks!

PR #16, opened 2020-12-15
Create README file to explain stack check demo

OscarShiang:
Make some description of how the stack checking mechanism works.

### DIFF
--- a/demos/stackcheck/README.md
+++ b/demos/stackcheck/README.md
@@ -1,0 +1,48 @@
+# Stack Check
+
+In this demo we demonstrate stack checking.
+
+Overflow occurs if the stack pointer exceeds the low address bound of the stack. The most common case is deep recursion. Since we only have limited space for the stack, we may clog the stack when we have too many function calls.
+
+In contrast, underflow is the situation that the stack pointer goes beyond the high address bound of the stack. We might encounter underflow when we return the first and the only function or the number of bytes we intend to write exceeds the high bound of the stack.
+
+It is hard to detect illegal use of memory and it may lead to the corruption of the kernel structure. So each time a thread does a syscall, we do checks to ensure the system can work properly.
+
+To illustrate the concept, we need to introduce the following:
+
+* `k_handle_syscall` (in `src/kernel/syscall.c`)
+* `check_stack` (in `src/kernel/syscall.c`)
+
+## `k_handle_syscall`
+
+This function is the entry when we call syscalls.
+
+Before we execute the syscall, it first checks whether the stack is overflowed/underflowed or not.
+
+If the stack is overflowed or underflowed, it will convert the syscall to `yield` to choose another valid thread.
+
+## `check_stack`
+
+This function is used to detect if the thread's stack has been overflowed or underflowed.
+
+In AMT, we place 2 variables `bottom_canary` and `top_canary` on both sides of the stack in the thread structure. These two variables are initialized to `STACK_CANARY`. If their values have been changed, we know that overflow/underflow has occurred.
+
+If the config `KCFG_DESTROY_ON_STACK_ERR` is enabled we invalidate the current thread and return `false`.
+
+If `KCFG_DESTROY_ON_STACK_ERR` is disabled, we just exit the kernel completely. (marking it as an unexpected exit)
+
+## Walkthrough
+
+In the beginning, `setup` thread enables `KCFG_DESTROY_ON_STACK_ERR` to keep the kernel alive when the overflow/underflow occurs.
+
+After setting up, the `setup` creates `watcher` and `underflow` thread and jumps into the `overflow()` function. When we jump into the `overflow()` function, it calls `recurse()` to produce an overflow.
+
+Since we keep calling `recurse()`, the value of the stack pointer will be repeatedly decreased to acquire stack space for these functions. Eventually, the stack will be overwhelmed and variables may be allocated to the address that does not belong to the stack. Thus when we revise the value of these variables, we may alter the structure of the thread. Overflow happens as a result.
+
+After the overflow occurs, the thread is not able to pass `check_stack()` so it will be marked as `INVALID_THREAD` when it tries to use syscalls. We'll yield to another thread instead.
+
+When we enter the `watcher` thread, it attempts to join `overflow` thread. Since the thread has been tagged as `INVALID_THREAD`, the `thread_join()` fails and returns `false`.
+
+After trying to join `underflow`, the thread yields to `underflow`. Since the `underflow` function is close to the top of the stack (to the bottom of the thread structure), when we write `distance` bytes from the address of `dummy`, we will overwrite the canary. After it yields, the kernel finds its stack is underflowed and then marks the thread as invalid.
+
+Finally, both `overflow` and `underflow` have been aborted, so the `watcher` fails to join these threads and ends the demo.
--- a/docs/demos.md
+++ b/docs/demos.md
@@ -14,7 +14,7 @@ For the best understanding we suggest that you:
 | [message](../demos/message/README.md)       | Passing messages between threads.                                                 |
 | [exit](../demos/exit/README.md)             | Threads exiting normally like any other C function.                               |
 | [spawn](../demos/spawn/README.md)           | One thread creating other threads.                                                |
-| stackcheck                                  | Detection of thread stack underflow or overflow when they try to yield.           |
+| [stackcheck](../demos/stackcheck/README.md) | Detection of thread stack underflow or overflow when they try to yield.           |
 | args                                        | Passing arguments to a thread.                                                    |
 | mutexes                                     | Locking a buffer using a mutex.                                                   |
 | timer                                       | Thread switching using a timer interrupt.                                         |


OscarShiang:
Btw, I think we can replace the checking of valid `tid` in `choose_next_thread()` to `check_stack()` since the latter has a more restricted checking on the thread structure.

OscarShiang:
I updated the commit to include the add of the link for this README in the demo doc.

demos/stackcheck/README.md
```diff
@@ -0,0 +1,37 @@
+# Stack Check
+
+Inside this demo, we demonstrate the usage of stack checking.
```
DavidSpickett: "In this demo we demonstrate stack checking."

Might be worth defining under and overflow as well. See: https://github.com/DavidSpickett/ARMMultiTasking/commit/498426d5c6ca752fdc1caeb40fdf173180859eaf

demos/stackcheck/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+It is hard to detect illegal usages of the memory, but it may lead to the corruption of the kernel structure. So we do checks to ensure the system can work properly after doing some operations like yielding.
```
DavidSpickett: "use of memory"
"and it may lead"

"So each time a thread yields, we do checks to ensure the system can work properly."

demos/stackcheck/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+To illustrate the concept, we need to introduce the following:
+* `check_stack` (in `src/kernel/thread.c`)
+* `choose_next_thread` (in `src/kernel/thread.c`)
+
+## `check_stack`
+
+This function is used to detect if the thread's stack has been overflowed or underflowed. 
+
+In AMT, we place 2 variables `bottom_canary` and `top_canary` on both sides of the stack in the thread structure. These two variables are initialized to `STACK_CANARY`. If their values have been changed, we can ensure that the overflow/underflow occurs.
```
DavidSpickett: "we know that overflow/underflow has occurred."

demos/stackcheck/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+## `choose_next_thread`
```
DavidSpickett: I think there's some confusion as to what check_stack does.

If the canaries have changed then...

If DESTROY_ON_STACK_ERR is enabled (meaning destroy the thread on stack err, the name isn't great) we invalidate the current thread then call load_next_thread.
next_thread is almost certainly null here but it could be non null (for a yield_to). So in a way yes, we do end up at choose_next_thread.

If DESTROY_ON_STACK_ERR isn't enabled then we just exit the kernel completely (marking it as an unexpected exit).

If I were you, I would devote this writeup to check_stack on its own and include the above detail.
You could add load_next_thread if you want, noting that it calls into assembly and will end up at the scheduler if it needs to.
But otherwise I don't think you need to describe anything else.

demos/stackcheck/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+We use this function to find the next thread to be run. If this function fails to find the next thread, it returns `NULL`.
+
+Before returning the address of the thread, it checks whether the thread has a right thread ID. If the `tid` differs from the index of the thread in `all_threads`, it implies that the thread structure has been destroyed. If the config `KCFG_DESTROY_ON_STACK_ERR` is disabled, the system will be aborted consequentially.
+
+## Walkthrough
+
+In the beginning, the `setup` thread enables `KCFG_DESTROY_ON_STACK_ERR` to keep the kernel alive when the overflow/underflow occurs. 
+
+After setting up, the `setup` creates `watcher` and `overflow` thread and jumps into the `underflow()` function. When we jump into the `underflow()` function, it calls `recurse()` to produce an underflow.
```
DavidSpickett: "After setting up, `setup` creates".

demos/stackcheck/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+Before returning the address of the thread, it checks whether the thread has a right thread ID. If the `tid` differs from the index of the thread in `all_threads`, it implies that the thread structure has been destroyed. If the config `KCFG_DESTROY_ON_STACK_ERR` is disabled, the system will be aborted consequentially.
+
+## Walkthrough
+
+In the beginning, the `setup` thread enables `KCFG_DESTROY_ON_STACK_ERR` to keep the kernel alive when the overflow/underflow occurs. 
+
+After setting up, the `setup` creates `watcher` and `overflow` thread and jumps into the `underflow()` function. When we jump into the `underflow()` function, it calls `recurse()` to produce an underflow. 
+
+Since we keep calling `recurse()`, the kernel will repeatedly allocate stack space to these functions. Eventually, the stack will be overwhelmed and variables may be allocated to the address that does not belong to the stack. Thus when we revise the value of these variables, we may alter the structure of the thread. Underflow happens as a result.
```
DavidSpickett: The kernel is not the one doing the allocation, this is static allocation. So the compiled code has the frame size baked in. Here's an example disassembly:
```
0000000000000dc0 <recurse>:
     dc0:       d108c3ff        sub     sp, sp, #0x230 << Each call to recurse uses this much space on the stack
     dc4:       a9007bfd        stp     x29, x30, [sp]
     dc8:       910003fd        mov     x29, sp
```

You have the rough idea though, keep allocating stack frames until the stack pointer ends up in the thread structure.

demos/stackcheck/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+After the underflow occurs, the thread is not able to pass `check_stack()` so it will be marked as `INVALID_THREAD` and yield to other threads.
+
+When we enter the `watcher` thread, it tries to join `underflowed` thread. Since the thread has been tagged as `INVALID_THREAD`, the `thread_join()` fails and returns `false`. 
+
+After trying to join `overflow`, the thread yields to `overflow`.  Since the variables in `overflow` thread are close to the bottom the thread structure, when we write `distance` bytes from the address of `dummy`, the overflow will appear. After it yields, the kernel finds its stack is overflowed and then terminates it.
```
DavidSpickett: "Since the `overflow` function is close to the top of the stack (to the bottom the thread structure), when we write `distance` bytes from the address of `dummy` we will overwrite the canary."

DavidSpickett (changes requested):
Once again this caused me to reread the demo and realise that over and underflow were backwards. (stacks growing down to get larger, it's confusing every time I think about it)

So I've reversed them and now it should make a bit more sense. Definitions in the commit I linked.

DavidSpickett:
> Btw, I think we can replace the checking of valid `tid` in `choose_next_thread()` to `check_stack()` since the latter has a more restricted checking on the thread structure.

I'm not sure what you mean, if you can post a diff for it that would clear things up.

OscarShiang:
The thing I said is to replace the checking on `tid` in `choose_next_thread()`. (like the diff below) 

```diff
diff --git a/src/kernel/thread.c b/src/kernel/thread.c
index 5ac5622..bcfbde1 100644
--- a/src/kernel/thread.c
+++ b/src/kernel/thread.c
@@ -268,12 +268,7 @@ static Thread* choose_next_thread(void) {
       continue;
     }

-    // We know that the ID is not -1 at this point
-    if ((size_t)all_threads[_idx].id != _idx) {
-      printf("thread ID %u and position %u inconsistent!\n", (unsigned)_idx,
-             all_threads[_idx].id);
-      k_exit(1);
-    }
+    check_stack();

     log_scheduler_event("scheduling new thread");
```

Since it checks if `tid` is consistent and I think that only stack overflow can cause `tid` to be inconsistent.
However, I didn't notice that `current_thread` might be `NULL` when loading the first thread.

I think I have misunderstood the meaning of checking the `tid`. Sorry about the weird comment I made.

OscarShiang:
I have fixed the README file.

OscarShiang:
I have fixed the README file and updated the commit.

DavidSpickett:
> The thing I said is to replace the checking on `tid` in `choose_next_thread()`. (like the diff below)
> 
> ```diff
> diff --git a/src/kernel/thread.c b/src/kernel/thread.c
> index 5ac5622..bcfbde1 100644
> --- a/src/kernel/thread.c
> +++ b/src/kernel/thread.c
> @@ -268,12 +268,7 @@ static Thread* choose_next_thread(void) {
>        continue;
>      }
> 
> -    // We know that the ID is not -1 at this point
> -    if ((size_t)all_threads[_idx].id != _idx) {
> -      printf("thread ID %u and position %u inconsistent!\n", (unsigned)_idx,
> -             all_threads[_idx].id);
> -      k_exit(1);
> -    }
> +    check_stack();
> 
>      log_scheduler_event("scheduling new thread");
> ```
> 
> Since it checks if `tid` is consistent and I think that only stack overflow can cause `tid` to be inconsistent.
> However, I didn't notice that `current_thread` might be `NULL` when loading the first thread.
> 
> I think I have misunderstood the meaning of checking the `tid`. Sorry about the weird comment I made.

No problem, it's a good observation. Yes in normal usage tid would only be inconsistent due to stack overflow.

You got me thinking about other points we could check the stack. Such as on syscall entry and direct yields (where the user just `svc`s instead of syscalling) or timer interrupts.

In general, without virtual memory this "protection" is never gonna be foolproof and is mostly to catch mistakes while developing the kernel. However we can certainly improve it, thanks for bringing that up.

I won't go do any of that for now though, let's get the demo docs in with the current state.

demos/stackcheck/README.md
```diff
@@ -0,0 +1,45 @@
...
+* `check_stack` (in `src/kernel/thread.c`)
+* `choose_next_thread` (in `src/kernel/thread.c`)
+
+## `check_stack`
+
+This function is used to detect if the thread's stack has been overflowed or underflowed. 
+
+In AMT, we place 2 variables `bottom_canary` and `top_canary` on both sides of the stack in the thread structure. These two variables are initialized to `STACK_CANARY`. If their values have been changed, we know that overflow/underflow has occurred.
+
+If the config `KCFG_DESTROY_ON_STACK_ERR` is enabled we invalid the current thread and load the next thread.
```
DavidSpickett: "invalidate the current"

demos/stackcheck/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+If `KCFG_DESTROY_ON_STACK_ERR` is disabled, we just exit the kernel completely. (marking it as an unexpected exit)
+
+## `choose_next_thread`
```
DavidSpickett: I don't think we should be describing choose_next_thread here.

I was reading it again and found the whole process could be simplified a lot so see https://github.com/DavidSpickett/ARMMultiTasking/commit/8cececfa3b9352f02c4cd7db3f82a23b93a01486 and https://github.com/DavidSpickett/ARMMultiTasking/commit/6e02b7d1f09ce6a9dee8d5d0a22c452f0c348f72 for the new state.

Basically:
* yield happens
* we do check_stack, which might invalidate the thread
* if we're doing a YIELD_ANY, just carry on. We'll never return to current if it had a bad stack
* if we're doing a YIELD_TO we have to be careful not to return to current if it has a bad stack and the yield target isn't valid

Also I removed the call into assembly load_next_thread, it was just confusing things.

Based on that new behavior I would think that k_do_yield is a candidate to be described instead of choose_next_thread. Let me know what you think, assuming the new behavior makes sense.

demos/stackcheck/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Walkthrough
+
+In the beginning, `setup` thread enables `KCFG_DESTROY_ON_STACK_ERR` to keep the kernel alive when the overflow/underflow occurs.
+
+After setting up, the `setup` creates `watcher` and `overflow` thread and jumps into the `overflow()` function. When we jump into the `overflow()` function, it calls `recurse()` to produce an underflow.
+
+Since we keep calling `recurse()`, the value of the stack pointer will be repeatedly decreased to acquire stack space for these functions. Eventually, the stack will be overwhelmed and variables may be allocated to the address that does not belong to the stack. Thus when we revise the value of these variables, we may alter the structure of the thread. Underflow happens as a result.
+
+After the underflow occurs, the thread is not able to pass `check_stack()` so it will be marked as `INVALID_THREAD` and yield to other threads.
```
DavidSpickett: "as `INVALID_THREAD` when it yields and we'll switch to another thread."

demos/stackcheck/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+When we enter the `watcher` thread, it tries to join `underflowed` thread. Since the thread has been tagged as `INVALID_THREAD`, the `thread_join()` fails and returns `false`.
+
+After trying to join `overflow`, the thread yields to `underflow`.  Since the `underflow` function is close to the top of the stack (to the bottom of the thread structure), when we write `distance` bytes from the address of `dummy`, we will overwrite the canary. After it yields, the kernel finds its stack is overflowed and then terminates it.
```
DavidSpickett: ". So"

(two spaces, remove one)

"then marks the thread invalid."

DavidSpickett (changes requested):

demos/stackcheck/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+Finally, both `overflow` and `underflow` have been aborted, so the `watcher` fails to join these threads and ends the demo.
```
DavidSpickett, line 48: Otherwise this walkthrough looks good.

I did change one of the yields to a yield_to, but this description still fits and we can assume they're looking at the demo source to. So I'd just keep this as is.

DavidSpickett:
I said I wasn't going to but, again, I can't help myself. I've applied check_stack to all syscalls in https://github.com/DavidSpickett/ARMMultiTasking/commit/a05151c4095b1608229aca6ac45ea0ffb0fca816.

Meaning that you could describe check_stack and k_handle_syscall (maybe just the beginning of it, as it relates to the stack checking) for the demo doc. With some small changes to the walkthrough.

One thread now syscalls instead of yielding, otherwise it's the same demo code.

This is much simpler. On every syscall, if the current thread has a stack error, ignore whatever it was doing and yield away from it.

OscarShiang:
I have force-pushed and fixed the README file.

OscarShiang:
I found that the system may be terminated when we underflow the stack with 0's. (like the diff)

```diff
diff --git a/demos/stackcheck/stackcheck.c b/demos/stackcheck/stackcheck.c
index e10165b..a4398f3 100644
--- a/demos/stackcheck/stackcheck.c
+++ b/demos/stackcheck/stackcheck.c
@@ -39,7 +39,7 @@ void underflow() {
      So underflow is garaunteed.
   */
   // Set INVALID_THREAD so scheduler doesn't check that position == ID
-  memset((void*)&dummy, INVALID_THREAD, distance);
+  memset((void*)&dummy, 0, distance);
   while (1) {
     yield();
   }
```

I think because we overwrite the 4th thread's id from `INVALID_THREAD` to `0` so `choose_next_thread` considers it is a valid thread while the `tid` is inconsistent.

Is it necessary to keep this demo has at most 4 threads?

demos/stackcheck/README.md
```diff
@@ -0,0 +1,55 @@
+# Stack Check
+
+In this demo we demonstrate stack checking.
+
+Overflow occurs if the stack pointer exceeds the low address bound of the stack. The most common case is deep recursion. Since we only have limited space for the stack, we may clog the stack when we have too many function calls.
+
+In contrast, underflow is the situation that the stack pointer goes beyond the high address bound of the stack. We might encounter underflow when we return the first and the only function or the number of bytes we intend to write exceeds the high bound of the stack.
+
+It is hard to detect illegal use of memory and it may lead to the corruption of the kernel structure. So each time a thread yields, we do checks to ensure the system can work properly.
```
DavidSpickett: "a thread does a syscall"

demos/stackcheck/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+This function is used to detect if the thread's stack has been overflowed or underflowed. 
+
+In AMT, we place 2 variables `bottom_canary` and `top_canary` on both sides of the stack in the thread structure. These two variables are initialized to `STACK_CANARY`. If their values have been changed, we know that overflow/underflow has occurred.
+
+If the config `KCFG_DESTROY_ON_STACK_ERR` is enabled we invalidate the current thread and return `false`.
+
+If `KCFG_DESTROY_ON_STACK_ERR` is disabled, we just exit the kernel completely. (marking it as an unexpected exit)
+
+## `choose_next_thread`
```
DavidSpickett: You can remove choose_next_thread. I don't see that it adds anything.

demos/stackcheck/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+In contrast, underflow is the situation that the stack pointer goes beyond the high address bound of the stack. We might encounter underflow when we return the first and the only function or the number of bytes we intend to write exceeds the high bound of the stack.
+
+It is hard to detect illegal use of memory and it may lead to the corruption of the kernel structure. So each time a thread yields, we do checks to ensure the system can work properly.
+
+To illustrate the concept, we need to introduce the following:
+
+* `k_handle_syscall` (in `src/kernel/syscall.c`)
+* `check_stack` (in `src/kernel/syscall.c`)
+* `choose_next_thread` (in `src/kernel/thread.c`)
```
DavidSpickett: Remove choose_next_thread

DavidSpickett (changes requested):

DavidSpickett:
> I found that the system may be terminated when we underflow the stack with 0's. (like the diff)
> 
> ```diff
> diff --git a/demos/stackcheck/stackcheck.c b/demos/stackcheck/stackcheck.c
> index e10165b..a4398f3 100644
> --- a/demos/stackcheck/stackcheck.c
> +++ b/demos/stackcheck/stackcheck.c
> @@ -39,7 +39,7 @@ void underflow() {
>       So underflow is garaunteed.
>    */
>    // Set INVALID_THREAD so scheduler doesn't check that position == ID
> -  memset((void*)&dummy, INVALID_THREAD, distance);
> +  memset((void*)&dummy, 0, distance);
>    while (1) {
>      yield();
>    }
> ```
> 
> I think because we overwrite the 4th thread's id from `INVALID_THREAD` to `0` so `choose_next_thread` considers it is a valid thread while the `tid` is inconsistent.
> 
> Is it necessary to keep this demo has at most 4 threads?

Yes you are correct, I got:
```
Thread     overflow: recursing
Thread     overflow: syscalling
Thread     <HIDDEN>: Stack overflow!
Thread      watcher: yielding
Thread    underflow: underflowing
Thread    underflow: yielding
Thread     <HIDDEN>: Stack underflow!
thread ID 3 and position 0 inconsistent!
```

With writing 0s.

If you move to 3 threads then the problem is that current_thread is right next to the last thread stack. (to underflow's stack)
```
$ aarch64-none-elf-nm stackcheck | sort
<...>
0000000040005ba0 B all_threads
0000000040005ba0 B _bstart
00000000400081c8 B current_thread
<...>
```

And we can see the underflow thread write to it:
```
#0  underflow ()
    at /home/david/projects/ARMMultiTasking/demos/stackcheck/stackcheck.c:34
34        size_t distance = (void*)(&dummy) - (void*)current_thread;
(gdb) watch current_thread
Hardware watchpoint 2: current_thread
(gdb) c
<...>
Hardware watchpoint 2: current_thread
Old value = (Thread *) 0x40007510 <all_threads+6512>
New value = (Thread *) 0x0 <_Reset>
<...>
(gdb) bt
#0  memset ()
    at /tmp/dgboter/bbs/rhev-vm1--rhe6x86_64/buildbot/rhe6x86_64--aarch64-none-elf/build/src/newlib-cygwin/newlib/libc/machine/aarch64/memset.S:148
#1  0x0000000000000ef8 in underflow () at /home/david/projects/ARMMultiTasking/demos/stackcheck/stackcheck.c:42
#2  0x0000000000003bc4 in thread_start () at /home/david/projects/ARMMultiTasking/src/kernel/thread.c:560
#3  0x0000000000000000 in ?? ()
```

From there we get stuck due to some null deference I bet, not really worth going into.

You could fix this particular situation by moving current_thread. However if you put it (and other critical vars) before the stacks then you risk overflow hitting them. I think this is the reason I left this comment in the kernel linker script:
```
    /* Thread structs go first! */
    *(.thread_structs)
    /* then the important zero init globals */
    *(.thread_vars_bss)
```
Since overflow was much more common when I was doing a lot of the initial development.
The point is that this is all pretty inexact. Without using hardware memory protection we won't catch everything.

So yes, it is needed to have an extra thread for the demo to give the underflow thread something to write into. Also for underflow to write INVALID_THREAD so that 4th thread doesn't trip the error.

But I didn't document that anywhere, so I will fix that. Thanks for raising it!

Edit: As for the overflow thread, that will be overflowing into kernel stack which is of course not being used while the thread is running. In any case the kernel starts from fresh on each entry. So bar some uninitialised variables the kernel code won't care what's on its stack before it starts.

Edit2: documented the thread number in https://github.com/DavidSpickett/ARMMultiTasking/commit/665732932142c212389c6a3526868ec75c790f41

OscarShiang:
OK, I have force-pushed again. 

In this commit, I removed the `choose_next_thread` statement and corrected the description of `overflow` thread.

DavidSpickett (approved):

DavidSpickett:
LGTM, thanks again!

PR #17, opened 2020-12-29
Add README file to args demo

OscarShiang:
Make some description of the args demo.

### DIFF
--- a/demos/args/README.md
+++ b/demos/args/README.md
@@ -1,0 +1,45 @@
+# Args
+
+In this demo we are going to show that we can create threads with a worker routine and pass at most 4 arguments to them.
+
+To describe the demo, we will introduce the following:
+
+* `k_add_thread` (in `src/kernel/thread.c`)
+* `thread_start` (in `src/kernel/thread.c`)
+
+## `k_add_thread`
+
+The functions like `add_named_thread`, `add_thread_from_worker`, and `add_named_thread_with_args` are mapped to syscall `add_thread`. So we will jump to this function when we call them.
+
+This function checks if the `args` we gave is `NULL`. If it is `NULL` the function creates a dummy `ThreadArgs` with 4 zeroed arguments then uses it as the new thread's arguments.
+
+## `thread_start`
+
+`thread_start` is the starting point of every thread, it loads the arguments from the thread structure and runs the worker routine.
+
+## Walkthrough
+
+For starters, `setup` thread creates named threads `printer` and `subprinter` with several arguments.
+
+The values we pass to `printer` and the corresponding arguments are listed below:
+
+* `repeat` = `3`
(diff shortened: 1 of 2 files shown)


demos/args/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+## `k_add_thread`
+
+The functions like `add_named_thread`, `add_thread_from_worker`, and `add_named_thread_with_args` are mapped to syscall `add_thread`. So we will jump to this function when we call them.
+
+This function checks if the `args` we gave is `NULL`. If it is `NULL` the function creates a dummy `ThreadArgs` with 4 zeroed arguments then uses it as the new thread's arguments.
+
+## `thread_start`
+
+`thread_start` is the starting point of every thread. When the scheduler chooses a thread to run, it loads the arguments from the thread structure and runs the worker routine.
```
DavidSpickett: This implies that the scheduler does the loading. I'd say:
"`thread_start` is the starting point of every thread, it loads the arguments from the thread structure and runs the worker routine."

demos/args/README.md
```diff
@@ -0,0 +1,45 @@
...
+## Walkthrough
+
+For starters, `setup` thread creates named threads `printer` and `subprinter` with several arguments.
+
+The values we pass to `printer` and the corresponding arguments are listed below:
+
+* `3` -> `repeat`
+* `"aardvark"` -> `phrase`
+* `2` -> `sub_printer`
+* `2` -> `start`
```
DavidSpickett: This would make more sense if we reverse the order. Like:

We pass the following arguments to `printer`:
* `repeat` = `3`

demos/args/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+After the creation, `setup` thread ends and yields. Then the scheduler chooses `printer` to run.
+
+Since the `printer` got `3` for `repeat`, it will print out `"aardvark"` 3 times. After that, it sends the value of `start` as a message to `subprinter` and ends.
+
+When `printer` finished its work, `subprinter` has been yielded and its arguments have been initialized as below:
```
DavidSpickett: I'm not sure you mean by "has been yielded".

Maybe:
"When `printer` finishes its work, the next thread will be`subprinter`."

demos/args/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+* `words` -> `words`
+* `4` -> `num_phrases`
+* `2` -> `offset`
```
DavidSpickett: Same comment as above

DavidSpickett (changes requested):
You should add a link to this file in the docs/demos file.

OscarShiang:
OK, I have fixed them.

DavidSpickett (approved):

PR #19, opened 2021-01-22
Use "fiber" for user space threads

jserv:
This patch replaced occurrences of "fibres" with "fiber", which is
commonly used for lightweight thread of execution in computer science.


README.md
```diff
@@ -24,7 +24,7 @@ For implementation details see the [design doc](docs/design.md).
 * Signal handling
 * Permissions and errno for error handling
 * Access to thread registers for tracing
-* User space threads (fibres)
+* User space threads (fiber)
```
DavidSpickett: "(fibers)" plural

include/port/port.h
```diff
@@ -47,7 +47,7 @@ size_t generic_semihosting_call(size_t operation, size_t* parameters);
 size_t generic_syscall(Syscall num, size_t arg1, size_t arg2, size_t arg3,
                        size_t arg4);
 
-// Used to catch returning fibres
+// Used to catch returning fiber
```
DavidSpickett: "returning fibers"

src/hw/arm_virt/yield.S
```diff
@@ -177,7 +177,7 @@ set_context:
   ldr r1,  [r0], #4 @ Our new PC
   bx r1
 
-@ Used to catch fibres returning normally
+@ Used to catch fiber returning normally
```
DavidSpickett: "a fiber" or "fibers"

src/hw/thumb_lm3s6965evb/yield.S
```diff
@@ -227,7 +227,7 @@ set_context:
   ldr r1,  [r0], #4 @ Our new PC
   bx r1
 
-@ Used to catch fibres returning normally
+@ Used to catch fiber returning normally
```
DavidSpickett: "a fiber" or "fibers"

DavidSpickett (changes requested):
My British spelling strikes again! Some plurals needed but otherwise looks good.

DavidSpickett:
Also you should rename user/include/fibre.h and user/fibre.c:
```
$ find -name "*fibre*"
./demos/fibres
./demos/fibres/fibres.c
./include/user/fibre.h
./src/user/fibre.c
```

DavidSpickett (changes requested):
The fibre.h (now fiber.h) include in the demo needs fixing. Which is my mistake because I forgot to enable actions on pull requests. Should be working now, so next update you'll get a full CI run to catch stuff like that.

I think there's one comment left using "fibre":
```
./src/user/fibre.c:  // Putting the address of the return context on the new fibre's stack
```

If you weren't already, I'm using these commands to look for the spelling:
```
$ grep -r fibre . --exclude-dir .git
$ find -name "*fibre*"
```

DavidSpickett (approved):
LGTM

You changed FiberContext too, I missed that myself.

PR #20, opened 2021-01-27
Add README for the condition variable

OscarShiang:
This commit includes the README file for condition variable and the modification of docs/demos.md.

### DIFF
--- a/demos/conditionvariables/README.md
+++ b/demos/conditionvariables/README.md
@@ -1,0 +1,50 @@
+# Condition Variables
+
+In this demo we explain the mechanism in condition variables.
+
+Condition variable is a synchronization tool which can block threads until a certain condition happens. 
+
+To show the concept behind it, we need to introduce the following:
+
+* `init_condition_variable` (in `src/user/condition_variable.c`)
+* `wait` (in `src/user/condition_variable.c`)
+* `signal` (in `src/user/condition_variable.c`)
+* `broadcast` (in `src/user/condition_variable.c`)
+
+## `init_condition_variable`
+
+This function initializes the condition variable pointed by `cv`.
+
+Before using functions like `wait()`, `signal()` and `broadcast()`, we must set up the variables with this function.
+
+## `wait`
+
+When we use this function, the thread will be blocked by the condition variable pointed by `cv`.
+
+After calling the function, the thread's state will be changed to `waiting`. This means it will remain idle until signalled through the same condition variable.
+
+## `signal`
+
+This function unblocks the first thread that is blocked on the specified condition variable `cv`.
+
+`signal()` unlocks the thread by setting the state to `suspended` so the thread can be chosen by the scheduler.
+
+If there was a waiting thread that was signalled, the function returns `true`. When there are no threads that can be signalled, it returns `false`.
+
+## `broadcast`
+
+This function unblocks all threads blocked on the specified condition variable `cv`.
+
+`broadcast()` has a similar functionality to `signal()`. `broadcast()` will wake up all the threads blocked by the condition variable while `signal()` only wakes up one of them.
+
+## Walkthrough
+
+The `setup()` function creates 5 threads that are set to run `waiter()` and initializes a condition variable called `cond_var`. After the creation, the `setup()` thread jumps to `signaller()` function.
+
+In `signaller()`, it first yields to check if the condition variable successfully blocks all other threads. Each `waiter` thread will print out `"Waiting..."` and then wait on `cond_var` thus we'll end up backing in the `signaller` thread. Then the `signaller` thread uses `signal()` to first wake and then yield to thread 1 and thread 2. The scheduler will choose the threads woken up by `cond_var` after the `signaller` thread yields.
+
+Once thread 1 and thread 2 have finished their own routine, the `signaller` thread calls `broadcast()` and wakes up all threads except itself. Then the `signaller` yields to the other threads that were signalled in the `broadcast()`, allowing them to finish. After that, the `signaller` tries to call `signal()`, but since there is no thread blocked by `cond_var`, `signal()` returns `false`.
+
+Then `signaller` creates another thread called `final_signal` and waits for `cond_var`, which shows that a thread can wait for a condition variable which has once been signalled.
+
+Then the `final_signal` thread calls `signal()` on `cond_var` and finishes. Next `signaller` runs, it has no work left to do so exits and ends the demo.
--- a/docs/demos.md
+++ b/docs/demos.md
@@ -7,28 +7,28 @@ For the best understanding we suggest that you:
 * make some changes to the demo to verify its behaviour
 * follow it in the debugger with `make debug_<demo>` (see [debugging.md](debugging.md) for more details)
 
-| Name                                        | Description                                                                       |
-| ------------------------------------------- | --------------------------------------------------------------------------------- |
-| [yielding](../demos/yielding/README.md)     | Threads yielding back to the scheduler.                                           |
-| [exyielding](../demos/exyielding/README.md) | Threads yielding directly to another thread or the next available thread.         |
-| [message](../demos/message/README.md)       | Passing messages between threads.                                                 |
-| [exit](../demos/exit/README.md)             | Threads exiting normally like any other C function.                               |
-| [spawn](../demos/spawn/README.md)           | One thread creating other threads.                                                |
-| [stackcheck](../demos/stackcheck/README.md) | Detection of thread stack underflow or overflow when they try to yield.           |
-| [args](../demos/args/README.md)             | Passing arguments to a thread.                                                    |
-| [mutexes](../demos/mutexes/README.md)       | Locking a buffer using a mutex.                                                   |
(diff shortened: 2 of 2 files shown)


demos/conditionvariables/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+This function initializes the condition variable pointed by `cv`.
+
+Before using functions like `wait()`, `signal()` and `broadcast()`, we must set up the variables with this function.
+
+## `wait`
+
+When we use this function, the thread will be blocked by the condition variable pointed by `cv`.
+
+After calling the function, the thread's `ThreadState` will be changed to `wait`. Since the scheduler only chooses the thread whose state is either `init`, `running` or `suspended`, the blocking thread will keep waiting until other threads signal through the same condition variable.
```
DavidSpickett: The state is `waiting`.

I would reword this.

"After calling the function, the thread's state will be changed to `waiting`. This means it will remain idle until signalled through the same condition variable."

That way we don't need to list all the states.

Side note: choosing a `running` thread *should* never happen but it's one of those things that probably could happen in some weird circumstance.

demos/conditionvariables/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## `signal`
+
+This function unblocks a thread that is blocked on the specified condition variable `cv`.
+
+`signal()` unlocks the thread by setting the state to `suspended`. So the thread can be chosen by the scheduler.
```
DavidSpickett: You can remove the full stop here, make it one sentence.

demos/conditionvariables/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+If there exists a thread that can be waked up, the function returns `true`. When there has no thread can be signalled, it returns `false` as a consequence.
```
DavidSpickett: The meaning here is fine but some phrasing changes would help.

"If there was a waiting thread that was signalled, the function returns `true`. When there are no threads that can be signalled, it returns `false`."

Side note: it would be "woken up" but I admit it also took me a while to work that one out :)

demos/conditionvariables/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## `wait`
+
+When we use this function, the thread will be blocked by the condition variable pointed by `cv`.
+
+After calling the function, the thread's `ThreadState` will be changed to `wait`. Since the scheduler only chooses the thread whose state is either `init`, `running` or `suspended`, the blocking thread will keep waiting until other threads signal through the same condition variable.
+
+## `signal`
+
+This function unblocks a thread that is blocked on the specified condition variable `cv`.
```
DavidSpickett: Clarify this some with:
"This function unblocks the first thread that is blocked on..."

This emphasises the difference between signal and broadcast which is later.

demos/conditionvariables/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+`signal()` unlocks the thread by setting the state to `suspended`. So the thread can be chosen by the scheduler.
+
+If there exists a thread that can be waked up, the function returns `true`. When there has no thread can be signalled, it returns `false` as a consequence.
+
+## `broadcast`
+
+This function unblocks all threads blocked on the specified condition variable `cv`.
+
+`broadcast()` has a similar functionality to `signal()`. However, when we use `broadcast()`, we will wake up all the threads blocked by the condition variable while `signal()` only wakes up one of them.
```
DavidSpickett: A little trim to the second sentence:
"`broadcast()` will wake up all the threads blocked by the condition variable while `signal()` only wakes up one of them."

demos/conditionvariables/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Walkthrough
+
+From starter, the `setup()` function creates 5 threads that are set to run `waiter()` and initializes a condition variable called `cond_var`. After the creation, the `setup()` thread jumps to `signaller()` function.
```
DavidSpickett: You could just start with:
"The `setup()` function..."

It's neater.

demos/conditionvariables/README.md
```diff
@@ -0,0 +1,50 @@
...
+If there exists a thread that can be waked up, the function returns `true`. When there has no thread can be signalled, it returns `false` as a consequence.
+
+## `broadcast`
+
+This function unblocks all threads blocked on the specified condition variable `cv`.
+
+`broadcast()` has a similar functionality to `signal()`. However, when we use `broadcast()`, we will wake up all the threads blocked by the condition variable while `signal()` only wakes up one of them.
+
+## Walkthrough
+
```
DavidSpickett, line 41: I see you're ignoring the misc. tests setup does at the very start. Which I think is a good way to go. 👍 

It's obvious they are kinda standalone things and they don't get in the way of the actual demo.

demos/conditionvariables/README.md
```diff
@@ -0,0 +1,50 @@
...
+From starter, the `setup()` function creates 5 threads that are set to run `waiter()` and initializes a condition variable called `cond_var`. After the creation, the `setup()` thread jumps to `signaller()` function.
+
+In `signaller()`, it first yields to check if the condition variable successfully blocks all other threads. Then the `signaller` thread uses `signal()` to wake 2 threads. The scheduler will choose the thread waked up by `cond_var` after the `signaller` thread yields.
```
DavidSpickett: "threads woken up by"

threads plural

Worth mentioning when the other threads do their wait. Something like:
"Each `waiter` thread will do some work and then wait on `cond_var` so we'll end up back in `signaller`. Then the `signaller` thread..."

I would be more clear what happens here. It's not that we signal 2 threads then yield and they run one after another, its signaller, thread, signaller, thread. (maybe that's what you're saying but it's a bit unclear).

Maybe:
"Then the `signaller` thread uses `signal()` to first wake and then yield to, 2 threads."

That makes the order clearer.

demos/conditionvariables/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+Once thread 1 and thread 2 finish their own routine, the `signaller` thread calls `broadcast()` and wakes up all threads except itself. After that, the `signaller` tries to call `signal()`, but since there is no thread blocked by `cond_var`, `signal()` returns `false`.
```
DavidSpickett: "Once thread 1 and thread 2 have finished"

I think "finished" makes it clearer that they have ended, not just got to a yield point.

demos/conditionvariables/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+Then `signaller` creates another thread called `final_signal` and waits for `cond_var`, which shows that a thread can wait for a condition variable which has once been signalled.
```
DavidSpickett, line 48: I would add:
"This wait causes a yield to the other threads that were signalled in the `broadcast`, allowing them to finish."

demos/conditionvariables/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+When the `final_signal` thread calls `signal()` on `cond_var`, it finishes its execution. Since the `signaller` has no job to do, it exits after it is chosen to run and ends the demo.
```
DavidSpickett: "Then the `final_signal` thread calls `signal()` on `cond_var` and finishes. Next `signaller` runs, it has no work left to do so exits and ends the demo."

DavidSpickett (changes requested):

demos/conditionvariables/README.md
OscarShiang: Isn't that the threads exit because of the `yield()` after we call `broadcast()`?

DavidSpickett: Doh, yes you're right!

demos/conditionvariables/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+This function unblocks all threads blocked on the specified condition variable `cv`.
+
+`broadcast()` has a similar functionality to `signal()`. `broadcast()` will wake up all the threads blocked by the condition variable while `signal()` only wakes up one of them.
+
+## Walkthrough
+
+From starter, the `setup()` function creates 5 threads that are set to run `waiter()` and initializes a condition variable called `cond_var`. After the creation, the `setup()` thread jumps to `signaller()` function.
+
+In `signaller()`, it first yields to check if the condition variable successfully blocks all other threads. Each `waiter` thread will print out `"Waiting..."` and then wait on `cond_var` thus we'll end up backing to the `signaller` thread. Then the `signaller` thread uses `signal()` to first wake and then yields to thread 1 and thread 2. The scheduler will choose the threads woken up by `cond_var` after the `signaller` thread yields.
```
DavidSpickett: "end up back in the `signaller` thread"

"wake and then yield to" (no "s" on the yield)

DavidSpickett (changes requested):

demos/conditionvariables/README.md
DavidSpickett: Ping on this in case it gets hidden by Github's UI.

DavidSpickett (changes requested):

DavidSpickett (approved):
LGTM

PR #21, opened 2021-01-27
Add description of cancel demo

OscarShiang:
Create a README file to explain the mechanism of cancel demo.

### DIFF
--- a/demos/cancel/README.md
+++ b/demos/cancel/README.md
@@ -1,0 +1,46 @@
+# Cancel
+
+In this demo we intend to show that we can use a syscall to terminate a thread and release the resources owned by it.
+
+To illustrate the mechanism, we must introduce the following:
+
+* `thread_wait` (in `src/user/thread.c`)
+* `thread_cancel` (in `src/user/thread.c`)
+
+## `thread_wait`
+
+This function sets the state of the current thread to `waiting` and yields.
+
+Since the scheduler will not pick a thread in `waiting` state, the thread will remain idle until another thread changes its state.
+
+## `thread_cancel`
+
+When we call this function, it will cancel the thread specified by the `tid`.
+
+If `tid` is given as `CURRENT_THREAD`, the current thread will be terminated. The scheduler will choose another thread to run.
+
+If we apply `cancel()` to a valid thread, the memory space it uses will be released. Then we can create another thread using the released memory space.
+
+When we call the function to cancel an invalid thread, it will return `false`. No thread will be terminated.
+
+## Walkthrough
+
+In the beginning, the `setup()` function creates a thread called `cancel_self` which cancels itself after it starts to run.
+
+When the scheduler chooses `setup` thread to run again, it creates another 3 threads.
+
+The `tid` of each thread and their corresponding routines are listed below:
+
+| tid | routine   | usage                  |
+| --- | --------- | ---------------------- |
+| 1   | work      | thread to be cancelled |
+| 2   | canceller | execute cancelling     |
+| 3   | work      | thread to be cancelled |
+
+The `setup` thread then sets its state to `waiting` and yields. The scheduler chooses thread 1 to run next. It prints out the message `"foo"` and yields.
+
+The `canceller` thread starts by checking that you can't cancel an invalid thread, then cancelling thread 3. Then `canceller` adds a new thread. We will place the new thread into space that used to store the thread we cancelled. As a result, the `tid` of the new thread will be `3`. We cancel the new thread 3 when we end up checking.
+
+After that, the `canceller` thread terminates thread 1. We see that its state is `cancelled` when we use `thread_join()` to join thread 1.
+
+Finally, the `canceller` cancels thread 0 (which is still in the `waiting` state) and returns. The demo now ends because there are no valid threads left.
--- a/docs/demos.md
+++ b/docs/demos.md
@@ -20,7 +20,7 @@ For the best understanding we suggest that you:
 | timer                                       | Thread switching using a timer interrupt.                                         |
 | threadlocalstorage (Arm/Thumb)              | Using thread local storage (TLS) to give each thread it's own 'global' variables. |
 | conditionvariables                          | Waiting on, signalling and broadcasting to condition variables.                   |
-| cancel                                      | Cancelling threads.                                                               |
+| [cancel](../demos/cancel/README.md)         | Cancelling threads.                                                               |
 | file                                        | Read from a file.                                                                 |
 | alloc                                       | Use of malloc/free.                                                               |
 | loadbinary                                  | Loading a thread from a separate binary.                                          |


demos/cancel/README.md
```diff
@@ -0,0 +1,36 @@
+# Cancel
+
+In this demo we intend to show that we can use a syscall to terminate a thread and release the resources that used to own by it.
```
DavidSpickett: "release the resources used by it." or "release the resources owned by it." whichever you prefer

demos/cancel/README.md
```diff
@@ -0,0 +1,36 @@
...
+To illustrate the mechanism, we must introduce the following:
+
+* `thread_wait` (in `src/user/thread.c`)
+* `thread_cancel` (in `src/user/thread.c`)
+
+## `thread_wait`
+
+This function sets the state of the current thread to `wait`.
+
+Since the scheduler will not pick a thread in `wait` state, the thread will keep blocking until we change its state.
```
DavidSpickett: The state value is `waiting` not `wait` (which also makes the explanation clearer)

Saying that it blocks isn't correct, but I see what you're going for. I would say:
"Since the scheduler will not pick a thread in `wait` state, the thread will remain idle until another thread changes its state."

demos/cancel/README.md
```diff
@@ -0,0 +1,36 @@
...
+In this demo we intend to show that we can use a syscall to terminate a thread and release the resources that used to own by it.
+
+To illustrate the mechanism, we must introduce the following:
+
+* `thread_wait` (in `src/user/thread.c`)
+* `thread_cancel` (in `src/user/thread.c`)
+
+## `thread_wait`
+
+This function sets the state of the current thread to `wait`.
```
DavidSpickett: "to wait and yields."

That's key here I think.

demos/cancel/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+Since the scheduler will not pick a thread in `wait` state, the thread will keep blocking until we change its state.
+
+## `thread_cancel`
+
+When we call this function, it will cancel the thread specified by the `tid`.
+
+If `tid` is given as `CURRENT_THREAD`, the current thread will be terminated. The scheduler will choose another thread to run.
+
+If we apply `cancel()` to a valid thread, the memory space used it used will be cleaned up. Then we can create another thread on the released memory space.
```
DavidSpickett: "the memory space it uses will be released."

I like the term "released" here, since the important point is that the memory is now free. Cleaned up can be a bit vague.

"Then we can create another thread using the released memory space."

I would swap "on" for "in" or, my suggestion "using". On also works but usually "in" is the term.

demos/cancel/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+When we call the function to cancel an invalid thread, it will return `false`. No thread will be terminated.
+
+## Walkthrough
+
+In the beginning, the `setup()` function creates a thread called `cancel_self` which cancels itself after it starts to run.
+
+When the scheduler chooses `setup` thread to run again, it creates another 3 threads and set itself to `wait` state. This is to occupy the memory space for the thread structure, which can ensure the operation later.
```
DavidSpickett: "sets its state to `waiting` and yields"

I think you can remove the "This is to occupy" sentence. Just reference it later instead.

I would state what each of the 3 new threads is doing and what ID they end up with. This will help later.

demos/cancel/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+Then `canceller` thread cancelled thread 3 and adds a new thread. Therefore we will place the new thread into space that used to store the cancelled thread before. As a result, the `tid` of the new thread will be `3`.
```
DavidSpickett: "The `canceller` thread starts by checking that you can't cancel an invalid thread, then cancelling thread 3."

Then you can describe adding the new thread.

demos/cancel/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+After that, the `canceller` thread terminated both thread 3 and thread 1. Since thread 1 has been cancelled, when we use `thread_join()` to join thread 1, we will see that its state is `cancelled`.
```
DavidSpickett: Here I would start with cancelling thread 1, cancelling 3 is covered above.

demos/cancel/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+Finally, the `canceller` cancels thread 0 and ends the demo.
```
DavidSpickett: Here you can reference that 0 is waiting.

"Finally, the `canceller` cancels thread 0 (which is still in the `waiting` state) and returns. The demo now ends because there are no valid threads left."

DavidSpickett (changes requested):

demos/cancel/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+The `tid` of each thread and their corresponding routines are listed below:
+
+| tid | routine   | usage                  |
+| --- | --------- | ---------------------- |
+| 1   | work      | thread to be cancelled |
+| 2   | canceller | execute cancelling     |
+| 3   | work      | thread to be cancelled |
+
+The `setup` thread then sets its state to `waiting` and yields.
```
DavidSpickett: I would add here that thread 1 does run once before the canceller starts.

demos/cancel/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+The `canceller` thread starts by checking that you can't cancel an invalid thread, then cancelling thread 3. The `canceller` adds a new thread. We will place the new thread into space that used to store the cancelled thread before. As a result, the `tid` of the new thread will be `3`.
```
DavidSpickett: This fine just some small corrections.

"Then `canceller` adds a new thread."
"used to store the thread we cancelled."

demos/cancel/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+After that, `canceller` thread terminated thread 1. We will see that its state is `cancelled` when we use `thread_join()` to join thread 1.
```
DavidSpickett: "After that, the `canceller` thread terminates thread 1. We see that it's state"

DavidSpickett (changes requested):

demos/cancel/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+| tid | routine   | usage                  |
+| --- | --------- | ---------------------- |
+| 1   | work      | thread to be cancelled |
+| 2   | canceller | execute cancelling     |
+| 3   | work      | thread to be cancelled |
+
+The `setup` thread then sets its state to `waiting` and yields. The scheduler chooses thread 1 to run next. It prints out the message `"foo"` and yields.
+
+The `canceller` thread starts by checking that you can't cancel an invalid thread, then cancelling thread 3. Then `canceller` adds a new thread. We will place the new thread into space that used to store the thread we cancelled. As a result, the `tid` of the new thread will be `3`.
```
DavidSpickett: Just add here that the new thread 3 is cancelled after we check that its ID is 3.

DavidSpickett (changes requested):

DavidSpickett (approved):
LGTM

PR #23, opened 2021-03-22
Make Arm semihosting support optional

OscarShiang:
In this patch, I add an new option in CMakeLists file to control whether
the semihosting feature should be turned on or off.

If semihosting has been disabled, the exit syscall will print out
"Exiting kernel" and hang in a while-loop. waiting the user to exit
manually.

Semihosting feature may not be provided by some devices. (e.g. Raspberry Pi)

I am currently working on porting AMT onto Raspberry Pi 4B, running 
semihosting may cause the board halted. So I think we should provide an extra
setting to control this feature.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,7 @@ set( OPT_LEVEL "3" CACHE STRING "Compiler optmisation level." )
 option( SANITIZERS "Enable UBSAN" OFF )
 option( LTO "Enable link time optimisation." OFF )
 option( CCACHE "Enable ccache." ON )
+set( SEMIHOSTING ON )
 set( STACK_SIZE 2 ) # In KB
 set( RAM_SIZE 0x100000 ) # 1MB in bytes
 
@@ -48,6 +49,7 @@ message(STATUS "SANITIZERS are ${SANITIZERS}")
 message(STATUS "LTO is ${LTO}")
 message(STATUS "COVERAGE is ${COVERAGE}")
 message(STATUS "STACK_SIZE is ${STACK_SIZE}KB")
+message(STATUS "SEMIHOSTING is ${SEMIHOSTING}")
 
 if(CCACHE)
   find_program(CCACHE_FOUND ccache)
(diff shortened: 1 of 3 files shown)


jserv (changes requested):
You can simplify the subject of this pull request to "Make Arm semihosting support optional."

OscarShiang:
OK, I will do it.

CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ set( OPT_LEVEL "3" CACHE STRING "Compiler optmisation level." )
 option( SANITIZERS "Enable UBSAN" OFF )
 option( LTO "Enable link time optimisation." OFF )
 option( CCACHE "Enable ccache." ON )
+option( SEMIHOSTING "Enable semihosting." ON )
```
DavidSpickett: If this is going to be set automatically per board (which is fine) it should be done like STACK_SIZE etc. Using set.
(option makes it seem like you can do cmake ... -DSEMIHOSTING=OFF, but you can't)

Also when you do that, you can remove the ON values for the current boards, just set it once, and over in your rpi fork just set OFF for that one.

src/hw/arm_common/port.c
```diff
@@ -27,6 +28,9 @@ size_t generic_semihosting_call(size_t operation, size_t* parameters) {
                : [ parameters ] "r"(parameters), [ operation ] "r"(operation)
                : RCHR "0", RCHR "1", "memory");
   return ret;
+#else
+  return -1;
```
DavidSpickett, line 34: Please mark the parameters used here. Like:
```
(void)operation; (void)parameters;
```

DavidSpickett (changes requested):

src/hw/arm_common/port.c
DavidSpickett: I'm going to enable -Wunused-params to catch this sort of thing.

OscarShiang:
OK, I have force-pushed the commit.

DavidSpickett (approved):
LGTM

PR #24, opened 2021-03-27
Create option of link script

OscarShiang:
In this patch, we create a new option in cmake to let us change link scripts.

Since there is a different address map on Raspberry Pi, I think we should provide such an option.
We can then replace the original one when we want to build the kernel for Raspberry Pi.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,7 @@ option( LTO "Enable link time optimisation." OFF )
 option( CCACHE "Enable ccache." ON )
 set( STACK_SIZE 2 ) # In KB
 set( RAM_SIZE 0x100000 ) # 1MB in bytes
+set( LINKER_SCRIPT "kernel.ld" )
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -9,6 +9,7 @@ option( LTO "Enable link time optimisation." OFF )
 option( CCACHE "Enable ccache." ON )
 set( STACK_SIZE 2 ) # In KB
 set( RAM_SIZE 0x100000 ) # 1MB in bytes
+set( LINK_SCRIPT "kernel.ld" )
```
DavidSpickett: LINKER_SCRIPT

DavidSpickett (changes requested):
Looks good, just change the option name.

OscarShiang:
I have fixed the option name in the new commit.

DavidSpickett (approved):

PR #26, opened 2021-05-18
Split the implementation of putchar from printf

OscarShiang:
The implementation of `putchar` may be various in different devices.

We use an platform-independent function `platform_putchar` to replace
the previous codes in `print.cpp`. So we can implement our own `putchar`
function in `src/hw/<specific_device>/port.c` instead of placing all
implementation in `print.cpp`.

In case that we need to initialize the device before using the platform
specific `putchar`, we also provide `platform_putc_init` function.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,7 @@ set( SEMIHOSTING ON )
 set( STACK_SIZE 2 ) # In KB
 set( RAM_SIZE 0x100000 ) # 1MB in bytes
 set( LINKER_SCRIPT "kernel.ld" )
+set( SERIAL_DRIVER "virt_serial.c" )
 
 string( TOLOWER "${BUILD_PLATFORM}" BP_LOWER )
 
@@ -19,7 +20,6 @@ if( BP_LOWER STREQUAL "arm" )
   set( PLATFORM     "-mcpu=cortex-a15" )
   set( PLATFORM_SRC "arm_virt" )
   set( RAM_START    "0x40000000" )
-  set( UART_BASE    "0x09000000" )
   set( QEMU_CMD     "qemu-system-arm -M virt -cpu cortex-a15" )
 elseif( BP_LOWER STREQUAL "aarch64" )
   set( PREFIX       "aarch64-none-elf-" )
@@ -27,7 +27,6 @@ elseif( BP_LOWER STREQUAL "aarch64" )
   set( PLATFORM     "-mcpu=cortex-a57 -mgeneral-regs-only" )
   set( PLATFORM_SRC "aarch64_virt" )
   set( RAM_START    "0x40000000" )
-  set( UART_BASE    "0x09000000" )
   set( QEMU_CMD     "qemu-system-aarch64 -M virt -cpu cortex-a57" )
   # O3 LTO UBSAN trace demo needs some extra
   set( STACK_SIZE 3 )
@@ -37,7 +36,6 @@ elseif( BP_LOWER STREQUAL "thumb" )
   set( PLATFORM_SRC "thumb_lm3s6965evb" )
   set( RAM_START    "0x20000000" )
   set( RAM_SIZE     0x10000)
-  set( UART_BASE    "0x4000C000" )
   set( QEMU_CMD     "qemu-system-arm -M lm3s6965evb -cpu cortex-m4" )
 else()
   message(FATAL_ERROR "Invalid platform \"${BP_LOWER}\". \
@@ -52,6 +50,7 @@ message(STATUS "COVERAGE is ${COVERAGE}")
 message(STATUS "STACK_SIZE is ${STACK_SIZE}KB")
 message(STATUS "SEMIHOSTING is ${SEMIHOSTING}")
 message(STATUS "LINKER_SCRIPT is ${LINKER_SCRIPT}")
+message(STATUS "SERIAL_DRIVER is ${SERIAL_DRIVER}")
 
 if(CCACHE)
   find_program(CCACHE_FOUND ccache)
@@ -85,7 +84,7 @@ endif()
 
 # -fno-common to disallow variables being present in multiple compliaton units
 set( CFLAGS "${CFLAGS} -ffreestanding -nostdlib -fno-common" )
-add_definitions(-DUART_BASE=${UART_BASE} -DSRC_ROOT=\"${CMAKE_SOURCE_DIR}\" )
+add_definitions( -DSRC_ROOT=\"${CMAKE_SOURCE_DIR}\" )
 
 if(LTO)
   set( CFLAGS "${CFLAGS} -flto -ffunction-sections" )
@@ -122,6 +121,8 @@ set( KERNEL_SOURCES
   src/hw/${PLATFORM_SRC}/startup.s
   src/hw/${PLATFORM_SRC}/yield.S
   src/hw/${PLATFORM_SRC}/port.c
+
+  src/hw/driver/${SERIAL_DRIVER}
 
   src/common/print.cpp
   src/common/trace.c
--- a/include/common/serial_port.h
+++ b/include/common/serial_port.h
@@ -1,0 +1,11 @@
+#ifndef COMMON_SERIAL_PORT
+#define COMMON_SERIAL_PORT
+
+typedef struct {
+  void (*init)(void);
+  void (*putchar)(int);
+} SerialPort;
+
+extern SerialPort serial_port;
+
+#endif /* COMMON_SERIAL_PORT */
--- a/src/common/print.cpp
+++ b/src/common/print.cpp
@@ -1,4 +1,5 @@
 #include "common/print.h"
+#include "common/serial_port.h"
 #include "common/thread.h"
 #include <ctype.h>
 #include <stdint.h>
@@ -13,6 +14,7 @@ extern "C" void __cxa_pure_virtual() {
 
 class PrintOutput {
 public:
+  PrintOutput() {}
   explicit PrintOutput(char* out) : m_out(out) {}
 
   virtual void write(int chr) const = 0;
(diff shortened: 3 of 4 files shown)


src/hw/aarch64_virt/port.c
```diff
@@ -22,3 +22,10 @@ void print_register_context(const RegisterContext* ctx) {
          ctx->x22, ctx->x23, ctx->x24, ctx->x25, ctx->x26, ctx->x27, ctx->x28,
          ctx->x29, ctx->x30, ctx->pc, ctx->spsr_el1, ctx->fpsr);
 }
+
+static volatile uint32_t* const UART0 = (uint32_t*)UART_BASE;
+
+void platform_putchar(int c) {
+  *UART0 = (uint32_t)c;
+  // We do not modify buf here, serial port doesn't move
```
jserv: The comment was misleading. Can you address the problems you encountered?

OscarShiang: This comment is used to address the difference between `SerialPrintOutput::write` and `BufferPrintOutput::write`.
But it actually becomes useless right here. So I will remove that.

include/port/port.h
```diff
@@ -50,4 +50,15 @@ size_t generic_syscall(Syscall num, size_t arg1, size_t arg2, size_t arg3,
 // Used to catch returning fibers
 void set_context_from_stack_address(void);
 
+#ifdef __cplusplus
```
DavidSpickett: Might as well put these around the whole file. Won't hurt and saves some head scratching if we reference more functions from C++.

include/port/port.h
```diff
@@ -50,4 +50,15 @@ size_t generic_syscall(Syscall num, size_t arg1, size_t arg2, size_t arg3,
...
+extern "C" {
+#endif
+
+void platform_putc_init();
```
DavidSpickett: Be consistent and call this `platform_putchar_init`.

DavidSpickett (changes requested):
I like the direction, I knew we'd end up with "drivers" of a sort eventually. This is exactly the kind of thing doing ports turns up!

That said I wonder if you can make it more like that model. What if we have `src/hw/driver/<name of the virt serial port>.c` and then (I'm guessing) later your `src/hw/driver/<name of rpi serial port>.c`?

Then we can de-duplicate the QEMU support bits.

Also it would be good to make a new header, like `include/common/serial_port.h`. Which includes the two functions you've added to port.h. That defines our generic serial port interface.

Then in cmake just conditionally add whichever driver C file fits the platform. If you forget to define it for a new port you'll get a linker error so I think it's pretty straightforward.

Oh and I merged your clang-format patch so CI should be good once you rebase this.

include/port/port.h
DavidSpickett: (Well, not if you don't add new the new functions here ofc)

OscarShiang:
I have added several changes in this new commit:
* **Add a new interface `SerialPrint`**:
    I add a new structure to specify the interface of serial printer.
    
    If one wants to replace the implementation from virt to another
    one, the things to do are to create an init function and a putchar
    function, and to create a `register_serial` to assign those two 
    functions to `platform_serial` object.

* **Create an option in CMakeLists.txt to control whether to add 
  `virt_serial.c` into kernel sources**:
    We can replace virt implementation by setting VIRT_DEVICE to OFF 
    and add the specific sources to kernel sources.

* **Replace the constructor of `init_threads` with `init_kernel_struct`**:                                                                                                                                   
    In case we will add more such interface intialization in, I use 
    a new function `init_kernel_struct` to run initializations like
    `init_threads` and `register_serial`.

OscarShiang:
It seems to have some problems when the executables were compiled with LTO enabled.

So I take `init_kernel_stuct` function out of the GNU constructors to avoid the errors.

include/port/port.h
```diff
@@ -50,4 +50,15 @@ size_t generic_syscall(Syscall num, size_t arg1, size_t arg2, size_t arg3,
...
+void platform_putchar(int c);
+
+#ifdef __cplusplus
+}
+#endif
+
```
DavidSpickett: This file doesn't need any changes now, right?

CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ set( SEMIHOSTING ON )
 set( STACK_SIZE 2 ) # In KB
 set( RAM_SIZE 0x100000 ) # 1MB in bytes
 set( LINKER_SCRIPT "kernel.ld" )
+set( VIRT_DRIVER ON )
```
DavidSpickett: I would rename this SERIAL_DRIVER since we've got more than one type now.

Then for your RPi stuff, you set SERIAL_DRIVER=src/hw/driver/<driver>.c

CMakeLists.txt
```diff
@@ -148,6 +150,10 @@ if(BP_LOWER STREQUAL "thumb" )
   set( KERNEL_SOURCES ${KERNEL_SOURCES} src/hw/${PLATFORM_SRC}/timer.c)
 endif()
 
+if(VIRT_DRIVER)
+  set( KERNEL_SOURCES ${KERNEL_SOURCES} src/hw/driver/virt_serial.c )
```
DavidSpickett: So here you would unconditionally add SERIAL_DRIVER to the sources.

include/common/serial_port.h
```diff
@@ -0,0 +1,13 @@
+#ifndef COMMON_SERIAL_PORT
+#define COMMON_SERIAL_PORT
+
+typedef struct {
+  void (*init)(void);
+  void (*putchar)(int);
+} SerialPrint;
```
DavidSpickett: SerialPrint -> SerialPort

include/common/serial_port.h
```diff
@@ -0,0 +1,13 @@
...
+
+extern SerialPrint platform_serial;
+
+void register_serial();
```
DavidSpickett: register_serial_port(void);

src/common/print.cpp
```diff
@@ -1,5 +1,7 @@
 #include "common/print.h"
+#include "common/serial_port.h"
 #include "common/thread.h"
+#include "port/port.h"
```
DavidSpickett: Probably not needed now.

include/common/serial_port.h
```diff
@@ -0,0 +1,13 @@
+#ifndef COMMON_SERIAL_PORT
+#define COMMON_SERIAL_PORT
+
+typedef struct {
+  void (*init)(void);
+  void (*putchar)(int);
+} SerialPrint;
+
+extern SerialPrint platform_serial;
```
DavidSpickett: I think it would be simpler to call this "serial_port".
```
extern SerialPort serial_port;
```

I know what you mean by platform but I don't think it's needed, we can see from context it's an abstraction.

src/hw/driver/virt_serial.c
```diff
@@ -0,0 +1,22 @@
+#include "common/serial_port.h"
+#include <stdint.h>
+
+SerialPrint platform_serial;
```
DavidSpickett: SerialPort serial_port;

src/hw/driver/virt_serial.c
```diff
@@ -0,0 +1,22 @@
...
+
+static volatile uint32_t* const UART0 = (uint32_t*)UART_BASE;
```
DavidSpickett, line 14: We could remove the cmake defined UART_BASE here by doing something like:
```
#ifdef __aarch64__
#define UART_BASE <the aarch64 value>;
#elif defined __thumb__
#define UART_BASE <the thumb value>
#elif defined __arm__
#define UART_BASE <the arm value>
#else
#error No virt serial port base address for this architecture!
#endif

static volatile uint32_t* const UART0 = (uint32_t*)UART_BASE;
```
Since that base is only relevant to virt.

src/hw/driver/virt_serial.c
```diff
@@ -0,0 +1,22 @@
...
+  *UART0 = (uint32_t)c;
+}
+
+void register_serial(void) {
+  SerialPrint virt = {
+      .init = virt_serial_init,
+      .putchar = virt_serial_putchar,
+  };
+  platform_serial = virt;
+}
```
DavidSpickett: I was able to simplify this by doing:
```
SerialPrint platform_serial = {
 .init = virt_serial_init,
 .putchar = virt_serial_putchar,
};
```
(after you define the functions obviously) Instead of having a register function we have to call, the global will just get init automatically.

src/kernel/thread.c
```diff
@@ -131,14 +132,19 @@ void init_thread(Thread* thread, int tid, const char* name,
   init_register_context(thread);
 }
 
-__attribute__((constructor)) static void init_threads(void) {
+static void init_threads(void) {
```
DavidSpickett: If you still need to mark something "constructor" can you elaborate on what went wrong with LTO?

I'm sure we can find an attribute that should/shouldn't be set to fix it.

However, I think if you init the serial struct as a global you won't need to make any changes to this file after all.

src/common/print.cpp
```diff
@@ -47,12 +49,12 @@ class PrintOutput {
 
 class SerialPrintOutput : public PrintOutput {
 public:
-  SerialPrintOutput() : PrintOutput(reinterpret_cast<char*>(UART_BASE)) {}
+  SerialPrintOutput() : PrintOutput(reinterpret_cast<char*>(UART_BASE)) {
+    platform_serial.init();
+  }
 
   void write(int chr) const final {
```
DavidSpickett, line 56: Also I just noticed these "final", I forgot C++ even had this feature. Good change.

DavidSpickett (changes requested):
This is looking great, exactly what I was thinking of.

Just a few things you can simplify.

OscarShiang:
OK, I have updated the commit.

DavidSpickett (approved):
LGTM

DavidSpickett:
I forgot to check that the commit message was updated, so I've just edited to summarise the changes as merged.

PR #27, opened 2021-05-21
Fix the format of `src/hw/arm_common/port.c`

OscarShiang:
The version of clang-format on GitHub Action seems to be higher
than the default version of Ubuntu 20.04. I had recently encountered
several errors of "Clang Format" check.

In this commit, I apply clang-format with version 13 and solve the error.


DavidSpickett:
I think this is due to us using `ubuntu-latest` which has now moved to 20.04, though I thought that had already happened. Anyway, that would give us clang (and therefore clang-format, I assume) 10.

(https://github.com/actions/virtual-environments/issues/1816)

This LGTM.

DavidSpickett:
Turns out it's clang-format-11, which is odd. It is installable from the llvm apt repo without being on 20.04 though so I'm gonna stick with it.

PR #28, opened 2021-06-25
Port BK onto RPi4B

OscarShiang:
In this commit, we add `raspi4` as a new build platform and create
a new category in `src/hw/` for raspi4-related implementations.

Since the bare metal way of reading/writing on SD card is still
working [in progress](https://www.raspberrypi.org/forums/viewtopic.php?f=72&t=308089), only the demos that are not related to file
system can be executed.

We also provides part of the functions in <string.h> to avoid
unaligned access from using toolchain-provided functions.

To test on your own Raspberry Pi 4, you can follow the steps in [Set up Raspberry Pi 4 to run AMT](https://hackmd.io/@oscarshiang/amt_setup).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,9 +32,11 @@ elseif( BP_LOWER STREQUAL "aarch64" )
   include(src/hw/aarch64_virt/CMakeLists.txt)
 elseif( BP_LOWER STREQUAL "thumb" )
   include(src/hw/thumb_lm3s6965evb/CMakeLists.txt)
+elseif( BP_LOWER STREQUAL "raspi4" )
+  include(src/hw/raspi4/CMakeLists.txt)
 else()
   message(FATAL_ERROR "Invalid platform \"${BP_LOWER}\". \
-  Expected one of \"arm\", \"thumb\", \"aarch64\".")
+  Expected one of \"arm\", \"thumb\", \"aarch64\", \"raspi4\".")
 endif()
 
 set( CMAKE_C_COMPILER "${PREFIX}gcc" )
--- a/cmake/AddDemo.cmake
+++ b/cmake/AddDemo.cmake
@@ -28,21 +28,23 @@ function(__add_demo NAME TEST_TYPE MAX_THREADS)
     VERBATIM)
   add_dependencies(make_demos ${NAME})
 
-  add_custom_target(run_${NAME})
-  add_dependencies(run_${NAME} ${NAME})
+  if (NOT BP_LOWER STREQUAL "raspi4")
+    add_custom_target(run_${NAME})
+    add_dependencies(run_${NAME} ${NAME})
 
-  add_custom_command(TARGET run_${NAME} POST_BUILD
-    COMMAND eval "${QEMU}${NAME}"
-   VERBATIM)
+    add_custom_command(TARGET run_${NAME} POST_BUILD
+      COMMAND eval "${QEMU}${NAME}"
+     VERBATIM)
 
-  add_custom_target(debug_${NAME})
-  add_dependencies(debug_${NAME} ${NAME})
+    add_custom_target(debug_${NAME})
+    add_dependencies(debug_${NAME} ${NAME})
 
-  add_custom_command(TARGET debug_${NAME} POST_BUILD
-    COMMAND eval "${QEMU}${NAME} -s -S"
-    VERBATIM)
+    add_custom_command(TARGET debug_${NAME} POST_BUILD
+      COMMAND eval "${QEMU}${NAME} -s -S"
+      VERBATIM)
+  endif()
 
-  if(NOT TEST_TYPE STREQUAL "none")
+  if(NOT TEST_TYPE STREQUAL "none" AND NOT BP_LOWER STREQUAL "raspi4")
     # This could be done with add_test, but then we wouldn't see the failure output.
     add_custom_target(test_${NAME} ALL)
     add_dependencies(test_${NAME} ${NAME})
@@ -67,5 +69,5 @@ function(__add_demo NAME TEST_TYPE MAX_THREADS)
     else()
       message(FATAL_ERROR "Unknown testing type \"${TEST_TYPE}\" for demo \"${NAME}\"")
     endif()
-  endif(NOT TEST_TYPE STREQUAL "none")
+  endif(NOT TEST_TYPE STREQUAL "none" AND NOT BP_LOWER STREQUAL "raspi4")
 endfunction(__add_demo)
--- a/cmake/AddLoadable.cmake
+++ b/cmake/AddLoadable.cmake
@@ -20,6 +20,8 @@ function(__add_loadable PARENT NAME PIE TEST)
   elseif( BP_LOWER STREQUAL "aarch64" )
     # Due to ardp alignment etc., we need more room
     set( CODE_PAGE_SIZE 11264 )
+  elseif( BP_LOWER STREQUAL "raspi4" )
+    set( CODE_PAGE_SIZE 11264 )
   else()
     message(FATAL_ERROR "Can't set CODE_PAGE_SIZE for \"${BP_LOWER}\".")
   endif()
--- a/include/port/raspi4/gpio.h
+++ b/include/port/raspi4/gpio.h
@@ -1,0 +1,22 @@
+#ifndef PORT_RASPI4_GPIO_H
+#define PORT_RASPI4_GPIO_H
+
+#include "port/raspi4/mmio.h"
+
+#define GPFSEL0 ((volatile unsigned int*)(MMIO_BASE + 0x00200000))
+#define GPFSEL1 ((volatile unsigned int*)(MMIO_BASE + 0x00200004))
+#define GPFSEL2 ((volatile unsigned int*)(MMIO_BASE + 0x00200008))
+#define GPFSEL3 ((volatile unsigned int*)(MMIO_BASE + 0x0020000C))
+#define GPFSEL4 ((volatile unsigned int*)(MMIO_BASE + 0x00200010))
+#define GPFSEL5 ((volatile unsigned int*)(MMIO_BASE + 0x00200014))
+#define GPSET0  ((volatile unsigned int*)(MMIO_BASE + 0x0020001C))
+#define GPSET1  ((volatile unsigned int*)(MMIO_BASE + 0x00200020))
+#define GPCLR0  ((volatile unsigned int*)(MMIO_BASE + 0x00200028))
+#define GPLEV0  ((volatile unsigned int*)(MMIO_BASE + 0x00200034))
+#define GPLEV1  ((volatile unsigned int*)(MMIO_BASE + 0x00200038))
+#define GPEDS0  ((volatile unsigned int*)(MMIO_BASE + 0x00200040))
+#define GPEDS1  ((volatile unsigned int*)(MMIO_BASE + 0x00200044))
+#define GPHEN0  ((volatile unsigned int*)(MMIO_BASE + 0x00200064))
+#define GPHEN1  ((volatile unsigned int*)(MMIO_BASE + 0x00200068))
+
+#endif /* PORT_RASPI4_GPIO_H */
--- a/include/port/raspi4/mbox.h
+++ b/include/port/raspi4/mbox.h
@@ -1,0 +1,26 @@
+#ifndef PORT_RASPI4_MBOX_H
+#define PORT_RASPI4_MBOX_H
+
+#include <stdint.h>
+
+#define MBOX_REQUEST 0
+
+/* channels */
+#define MBOX_CH_POWER 0
+#define MBOX_CH_FB    1
+#define MBOX_CH_VUART 2
+#define MBOX_CH_VCHIQ 3
+#define MBOX_CH_LEDS  4
+#define MBOX_CH_BTNS  5
+#define MBOX_CH_TOUCH 6
+#define MBOX_CH_COUNT 7
+#define MBOX_CH_PROP  8
+
+/* tags */
+#define MBOX_TAG_SETPOWER   0x28001
+#define MBOX_TAG_SETCLKRATE 0x38002
+#define MBOX_TAG_LAST       0
+
+int mbox_call(unsigned char ch, const uint32_t* mbox);
+
+#endif /* PORT_RASPI4_MBOX_H */
--- a/linker/kernel_rpi.ld
+++ b/linker/kernel_rpi.ld
@@ -1,0 +1,47 @@
+ENTRY(_Reset)
+
+SECTIONS {
+ . = 0x80000;
+
+ .text : {
+   KEEP(*(.text.boot))
+   *(.text*)
+   *(.rodata*)
+   . = ALIGN(8); /* align to pointer size (8 for AArch64) */
+   _init_array = . ;
+   *(.init_array*)
+   _einit_array = . ;
+   _etext = . ;
+ }
+
+ .data : {
+   _data = . ;
+   *(.data)
+   *(.thread_vars)
+   _edata = . ;
+ }
+
+ . = ALIGN(16); /* align 16 for AArch64 */
+ . = . + 0xC00; /* 3kB of stack for kernel */
+ stack_top = . ;
+
+ .bss (NOLOAD) : {
+   _bstart = . ;
+   /* Thread structs go first! */
+   *(.thread_structs)
+   /* then the important zero init globals */
+   *(.thread_vars_bss)
+   *(.bss COMMON)
+   /* Putting this first minimises waste on Arm/Thumb due to alignment below */
+   *(.code_page_backing)
+   /* AArch64 PIE requires that programs start on 4k alignemnt for adrp usage. */
+   . = ALIGN(0x1000);
+   /* on the end in case of weird stuff writing beyond */
+   *(.code_page)
+   . = ALIGN(16);
+   _bend = . ;
+ }
+
+ /* Don't need unwinding info on Arm */
+ /DISCARD/ : { *(.ARM.exidx* ) }
+}
--- a/src/hw/raspi4/CMakeLists.txt
+++ b/src/hw/raspi4/CMakeLists.txt
@@ -1,0 +1,22 @@
+set( KERNEL_SOURCES ${KERNEL_SOURCES}
+  ${CMAKE_SOURCE_DIR}/src/hw/arm_common/port.c
+  ${CMAKE_SOURCE_DIR}/src/hw/arm_common/gic.c
+  ${CMAKE_SOURCE_DIR}/src/hw/driver/raspi4_serial.c
+  ${CMAKE_SOURCE_DIR}/src/hw/raspi4/port.c
+  ${CMAKE_SOURCE_DIR}/src/hw/raspi4/startup.s
+  ${CMAKE_SOURCE_DIR}/src/hw/raspi4/yield.S
+  ${CMAKE_SOURCE_DIR}/src/hw/raspi4/yield.S
+  ${CMAKE_SOURCE_DIR}/src/hw/raspi4/uart.c
+  ${CMAKE_SOURCE_DIR}/src/hw/raspi4/mbox.c
+  ${CMAKE_SOURCE_DIR}/src/hw/raspi4/exception.c
+  ${CMAKE_SOURCE_DIR}/src/hw/raspi4/libs/string.c
+)
+
+set( PREFIX       "aarch64-none-elf-" )
+# -mgerneal-regs-only to generate instructions without using NEON regs
+# -mstrict-align to avoid unaligned memory access
+set( PLATFORM     "-mcpu=cortex-a72 -mgeneral-regs-only -mstrict-align" )
+# O3 LTO UBSAN trace demo needs some extra
+set( STACK_SIZE 3 )
+set( LINKER_SCRIPT "kernel_rpi.ld" )
+set( SEMIHOSTING OFF )
--- a/src/hw/raspi4/exception.c
+++ b/src/hw/raspi4/exception.c
@@ -1,0 +1,107 @@
+#include "port/raspi4/uart.h"
+
+void invalid_entry() {
+  // In case the UART isn't initialized
+  uart_init();
+  uart_puts("[Error] Invalid exception occurred\n");
+
+  // Load and decode the exception
+  unsigned long esr, elr, spsr, far;
+  asm volatile(
+      "mrs %[esr], esr_el1\n\t"
+      "mrs %[elr], elr_el1\n\t"
+      "mrs %[spsr], spsr_el1\n\t"
+      "mrs %[far], far_el1\n\t"
+      : [esr] "=r"(esr), [elr] "=r"(elr), [spsr] "=r"(spsr), [far] "=r"(far)::);
+
+  // Decode exception type (some, not all. See ARM DDI0487G_a chapter D13.2.37)
+  switch (esr >> 26) {
+    case 0b000000:
+      uart_puts("Unknown");
+      break;
+    case 0b000001:
+      uart_puts("Trapped WFI/WFE");
+      break;
+    case 0b001110:
+      uart_puts("Illegal execution");
+      break;
+    case 0b010101:
+      uart_puts("System call");
+      break;
+    case 0b100000:
+      uart_puts("Instruction abort, lower EL");
+      break;
+    case 0b100001:
+      uart_puts("Instruction abort, same EL");
+      break;
+    case 0b100010:
+      uart_puts("Instruction alignment fault");
+      break;
+    case 0b100100:
+      uart_puts("Data abort, lower EL");
+      break;
+    case 0b100101:
+      uart_puts("Data abort, same EL");
+      break;
+    case 0b100110:
+      uart_puts("Stack alignment fault");
+      break;
+    case 0b101100:
+      uart_puts("Floating point");
+      break;
+    default:
+      uart_puts("Unknown");
+      break;
+  }
+  // Decode data abort cause
+  if (esr >> 26 == 0b100100 || esr >> 26 == 0b100101) {
+    uart_puts(", ");
+    switch ((esr >> 2) & 0x3) {
+      case 0:
+        uart_puts("Address size fault");
+        break;
+      case 1:
+        uart_puts("Translation fault");
+        break;
+      case 2:
+        uart_puts("Access flag fault");
+        break;
+      case 3:
+        uart_puts("Permission fault");
+        break;
+    }
+    switch (esr & 0x3) {
+      case 0:
+        uart_puts(" at level 0");
+        break;
+      case 1:
+        uart_puts(" at level 1");
+        break;
+      case 2:
+        uart_puts(" at level 2");
+        break;
+      case 3:
+        uart_puts(" at level 3");
+        break;
+    }
+  }
+
+  // Dump registers
+  uart_puts(":\n  ESR_EL1 ");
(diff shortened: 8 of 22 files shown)


CMakeLists.txt
```diff
@@ -37,6 +37,19 @@ elseif( BP_LOWER STREQUAL "thumb" )
...
+elseif( BP_LOWER STREQUAL "raspi4" )
+  set( PREFIX       "aarch64-none-elf-" )
+  # Don't generate NEON
+  set( PLATFORM     "-mcpu=cortex-a72 -mgeneral-regs-only" )
+  set( PLATFORM_SRC "raspi4" )
+  set( QEMU_CMD     "qemu-system-aarch64 -M virt -cpu cortex-a57" )
+  # O3 LTO UBSAN trace demo needs some extra
+  set( STACK_SIZE 3 )
+  # Define here to not generate error when expanding macros
+  set( RAM_START    "0x0" )
```
DavidSpickett: I've just pushed a commit that means you shouldn't have to set this. I've moved all these values into platform specific linker scripts.

(using your LINKER_SCRIPT option, thanks again for that!)

CMakeLists.txt
```diff
@@ -149,6 +162,14 @@ if(BP_LOWER STREQUAL "thumb" )
   set( KERNEL_SOURCES ${KERNEL_SOURCES} src/hw/${PLATFORM_SRC}/timer.c)
 endif()
 
+if(BP_LOWER STREQUAL "raspi4" )
+  set( KERNEL_SOURCES ${KERNEL_SOURCES} src/hw/${PLATFORM_SRC}/uart.c)
+  set( KERNEL_SOURCES ${KERNEL_SOURCES} src/hw/${PLATFORM_SRC}/mbox.c)
+  set( KERNEL_SOURCES ${KERNEL_SOURCES} src/hw/${PLATFORM_SRC}/exception.c)
+  # Avoid from using NEON related implementations
```
DavidSpickett: I think that `-mgeneral-regs-only` should take care of that so really the reason is to avoid this unaligned access right? The comment isn't needed.

On that subject, I think you should add the gcc option `-mstrict-align`. Compiling this file now might not generate unaligned accesses but it doesn't seem like anything is stopping it from doing so if the compiler decided to. That option should fix that.

Also can you check if adding that option allows you to use the toolchain provided functions? It might not but if we're lucky it'll effect library selection as well, and we won't need the separate file.

cmake/AddLoadable.cmake
```diff
@@ -20,6 +20,8 @@ function(__add_loadable PARENT NAME PIE TEST)
   elseif( BP_LOWER STREQUAL "aarch64" )
     # Due to ardp alignment etc., we need more room
     set( CODE_PAGE_SIZE 11264 )
+  elseif( BP_LOWER STREQUAL "raspi4" )
```
DavidSpickett, line 23: Seeing as loadable demos would be disabled, you shouldn't need to set this.

If you get errors because the loadable demos are still being added, you should change the demo's CMakeLists.txt to require SEMIHOSTING. (which is a good change in general anyway)

include/port/raspi4/gpio.h
```diff
@@ -0,0 +1,22 @@
+#ifndef PORT_RASPI4_GPIO_H
```
DavidSpickett, line 1: Given that the GPIO defines here aren't used yet, I would rename this mmio.h and only include the MMIO_BASE define.

Then later you can add gpio.h that includes that. A bit silly to have a whole file for 1 define but I think it will help the organisation later.

include/port/raspi4/mbox.h
```diff
@@ -0,0 +1,27 @@
+#ifndef PORT_RASPI4_MBOX_H
+#define PORT_RASPI4_MBOX_H
+
+/* a properly aligned buffer */
```
DavidSpickett: You can remove this comment, given that the alignment is described in the C file.

In fact, if you added a function like `mbox_write(unsigned int* data, size_t len)` then you wouldn't need to expose this variable outside mbox.c.

(for the error handling for that just throw in some asserts that data is not null and len is < 36)

linker/kernel_rpi.ld
```diff
@@ -0,0 +1,44 @@
+ENTRY(_Reset)
```
DavidSpickett, line 1: See the recent changes to linker scripts, please format it in that style.

src/hw/raspi4/exception.c
```diff
@@ -0,0 +1,9 @@
+#include "port/raspi4/uart.h"
+
+void invalid_entry() {
+  uart_init();
+  uart_puts("[Error] Invalid exception occurs\n");
```
DavidSpickett: "Invalid exception occurred"

src/hw/raspi4/exception.c
```diff
@@ -0,0 +1,9 @@
+#include "port/raspi4/uart.h"
+
+void invalid_entry() {
+  uart_init();
```
DavidSpickett, line 5: Either remove this because it's not needed, or comment why it is. (the call to init)

src/hw/raspi4/libs/string.c
```diff
@@ -0,0 +1,44 @@
+#include <string.h>
+
```
DavidSpickett, line 10: Comment here why we need separate copies.

(assuming that flag I mentioned doesn't allow you to use the library versions)

src/hw/raspi4/mbox.c
```diff
@@ -0,0 +1,42 @@
...
+#define MBOX_SENDER    ((volatile unsigned int*)(VIDEOCORE_MBOX + 0x14))
+#define MBOX_STATUS    ((volatile unsigned int*)(VIDEOCORE_MBOX + 0x18))
+#define MBOX_CONFIG    ((volatile unsigned int*)(VIDEOCORE_MBOX + 0x1C))
+#define MBOX_WRITE     ((volatile unsigned int*)(VIDEOCORE_MBOX + 0x20))
+#define MBOX_RESPONSE  0x80000000
+#define MBOX_FULL      0x80000000
+#define MBOX_EMPTY     0x40000000
+
+int mbox_call(unsigned char ch) {
+  unsigned int r = (((unsigned int)((unsigned long)&mbox) & ~0xF) | (ch & 0xF));
```
DavidSpickett: Please add a comment explaining this line.

Add an assert that `ch <= 0xf`.

Why the cast to unsigned long then to int again? Consider using uintptr_t instead or just cast into unsigned int which will be 4 bytes, the same size as a pointer anyway.

If we have already 16 byte aligned the address of mbox, there's no need to remove the bottom 4 bits of &mbox.

src/hw/raspi4/mbox.c
```diff
@@ -0,0 +1,42 @@
...
+
+  /* now wait for the response */
+  while (1) {
+    /* is there a response? */
+    do {
+      asm volatile("nop");
+    } while (*MBOX_STATUS & MBOX_EMPTY);
+
+    /* is it a response to our message? */
+    if (r == *MBOX_READ)
```
DavidSpickett: Add braces {} here.

I know it's valid syntax but I prefer anything with > 1 line, even a comment line, to have braces.

src/hw/raspi4/startup.s
```diff
@@ -0,0 +1,97 @@
+.section ".text.boot"
+
+.globl _Reset
+_Reset:
+  // read cpu id, stop slave cores
```
DavidSpickett: Avoid the term `slave` here, just say "Stop every core that isn't core 0".

src/hw/raspi4/startup.s
```diff
@@ -0,0 +1,97 @@
...
+  mrs x1, mpidr_el1
+  and x1, x1, #3
+  cbz x1, 2f
+  // cpu id > 0, stop
+1:wfe
```
DavidSpickett: When using labels please name them, even if the names are seemingly super obvious.

Branching to numbers with the 1f kinda syntax is usually used when names will get repeated. If you can give them textual names do so.

DavidSpickett: Also put labels on their own line. For example here I'd do:
```
wfe:
  wfe
  b wfe
on_core_0:
 <...>
```

src/hw/raspi4/startup.s
```diff
@@ -0,0 +1,97 @@
...
+  b 1b
+2:// cpu id == 0
+  
+  // set stack before our code
```
DavidSpickett: I see what you mean but the wording is weird.

"Set initial stack pointer" ?

src/hw/raspi4/startup.s
```diff
@@ -0,0 +1,97 @@
...
+  ldr x1, =stack_top
+  
+  // set up EL1
```
DavidSpickett: Again what do you mean here? Is this sequence dropping *into* EL1?

src/hw/raspi4/startup.s
```diff
@@ -0,0 +1,97 @@
...
+  mrs x0, CurrentEL
+  and x0, x0, #12 // clear reserved bits
+  
+  // running at EL3?
+  cmp x0, #12
+  bne 5f
+  // should never be executed, just for completeness
```
DavidSpickett: Explain why we don't expect to see this, and what the code below *would* do if it were to happen.

src/hw/raspi4/startup.s
```diff
@@ -0,0 +1,97 @@
...
+  mov x2, #0x5b1
+  msr scr_el3, x2
+  mov x2, #0x3c9
+  msr spsr_el3, x2
+  adr x2, 5f
+  msr elr_el3, x2
+  eret
+  
+  // running at EL2?
+5:cmp x0, #4
```
DavidSpickett: Please space this section out, I'd suggest a newline before each comment. Maybe 3/4 distinct blocks of code.

src/hw/raspi4/startup.s
```diff
@@ -0,0 +1,97 @@
...
+  // enable CNTP for EL1
+  mrs x0, cnthctl_el2
+  orr x0, x0, #3
+  msr cnthctl_el2, x0
+  msr cntvoff_el2, xzr
+  // enable AArch64 in EL1
+  mov x0, #(1 << 31) // AArch64
+  msr hcr_el2, x0
+  mrs x0, hcr_el2
+  // enable FP / SIMD
```
DavidSpickett: Hang on, I thought you didn't want Neon?

One of the reasons we don't enable neon is that neon registers aren't saved and restored by the thread switching mechanism. So I would invert this to explicitly *disable* Neon.
(assuming that doesn't trip up some library code but in that case leave it as is, at least any compiled code will be fine)

src/hw/raspi4/startup.s
```diff
@@ -0,0 +1,97 @@
...
+  bl entry
+  // for failsafe, halt this core too
+  b 1b
+
+  .macro ventry label
+    .align 7
+    b \label
+  .endm
+
+.align 11
```
DavidSpickett, line 99: 11? I know the tables do have to be aligned but 11 seems a bit random

src/hw/raspi4/uart.c
```diff
@@ -0,0 +1,76 @@
...
+#define UART0_FR   ((volatile unsigned int*)(MMIO_BASE + 0x00201018))
+#define UART0_IBRD ((volatile unsigned int*)(MMIO_BASE + 0x00201024))
+#define UART0_FBRD ((volatile unsigned int*)(MMIO_BASE + 0x00201028))
+#define UART0_LCRH ((volatile unsigned int*)(MMIO_BASE + 0x0020102C))
+#define UART0_CR   ((volatile unsigned int*)(MMIO_BASE + 0x00201030))
+#define UART0_IMSC ((volatile unsigned int*)(MMIO_BASE + 0x00201038))
+#define UART0_ICR  ((volatile unsigned int*)(MMIO_BASE + 0x00201044))
+
+void uart_init() {
+  register unsigned int r;
```
DavidSpickett: `register` seems unneeded.

If it's needed to point codegen in the right direction for some reason, I'd prefer explicit use of inline asm. Or at least a good comment why it's needed.

src/hw/raspi4/uart.c
```diff
@@ -0,0 +1,76 @@
...
+
+  /* initialize UART */
+  *UART0_CR = 0; // turn off UART0
+
+  /* set up clock for consistent divisor values */
+  mbox[0] = 9 * 4;
```
DavidSpickett: Please comment all these magic numbers with their origin.

E.g. (making this up but...)

mbox[0] = 9 * 4; // 9 whatevers and 4 magic unicorns

You the idea :)

src/hw/raspi4/uart.c
```diff
@@ -0,0 +1,76 @@
...
+  mbox[4] = 8;
+  mbox[5] = 2;       // UART clock
+  mbox[6] = 4000000; // 4Mhz
+  mbox[7] = 0;       // clear turbo
+  mbox[8] = MBOX_TAG_LAST;
+  mbox_call(MBOX_CH_PROP);
+
+  /* map UART0 to GPIO pins */
+  r = *GPFSEL1;
+  r &= ~((7 << 12) | (7 << 15)); // gpio14, gpio15
```
DavidSpickett: // gpio14, gpio15

But the numbers shift *12* and 15.

Intentional?

src/hw/raspi4/vectors.s
```diff
@@ -0,0 +1 @@
+/* Empty file to keep the build system happy. */
```
DavidSpickett: The way our cmake setup is done it's very "top level file knows about everything".

The proper way is to have subdirs have their own cmakelists and they can decide what source files get added. But that's not relevant right now, just a thing I've been thinking about.

src/hw/raspi4/yield.S
```diff
@@ -0,0 +1,267 @@
...
+  // Store current sp/lr (used as the new pc)
+  stp x3, x30, [x0], #16
+  // Now we can set the new lr
+  mov x30, x4
+  // Set new sp
+  mov sp, x1
+
+  // Reset x0 to be arg1 of new function
+  sub x0, x0, #14*8
+  br x2
```
DavidSpickett, line 267: I'm assuming this is an exact copy of the aarch64_virt file (which is fine, code sharing is awkward with the current structure).

Please make sure that it is identical, if not note any divergences.

DavidSpickett (changes requested):
Exciting stuff!

DavidSpickett:
Also it would be cool to get the document you linked under `docs/` but I assume you're still finishing that so don't need to add it in this change.

src/hw/raspi4/startup.s
OscarShiang: It is because we will access `FPSR` when we store / load the CPU context.
We would be trapped if we did not enable that.

I think maybe I will add more comments about that.

CMakeLists.txt
```diff
@@ -37,6 +37,19 @@ elseif( BP_LOWER STREQUAL "thumb" )
...
+  set( QEMU_CMD     "qemu-system-aarch64 -M virt -cpu cortex-a57" )
+  # O3 LTO UBSAN trace demo needs some extra
+  set( STACK_SIZE 3 )
+  # Define here to not generate error when expanding macros
+  set( RAM_START    "0x0" )
+  set( SEMIHOSTING OFF )
+  set( LINKER_SCRIPT "kernel_rpi.ld" )
+  set( SERIAL_DRIVER "raspi4_serial.c" )
 else()
   message(FATAL_ERROR "Invalid platform \"${BP_LOWER}\". \
```
DavidSpickett: Add raspi4 to this error message

CMakeLists.txt
OscarShiang: I think that `-mgeneral-regs-only` only helps us not to generate NEON code on our own code.
But the toolchain implementation is already compiled. 
So we will be linked to some NEON-related implementation even if we enable this option.

DavidSpickett:
Also I'd like to setup CI so we can at least build this port. But let's make that a different change.

CMakeLists.txt
OscarShiang: I have not tried `-mstrict-align` before. I will give it a try.

DavidSpickett:
In the meantime I've made some cmake changes. The short version is that you should now put RPI related settings in src/hw/raspi4/CMakeLists.txt and all you have to do in the top level is extend the platform if/else.

src/hw/raspi4/vectors.s
DavidSpickett: I've fixed this now, you can choose which files to add so you don't need this dummy file. See src/hw/aarch64_virt/CMakeLists.txt for how to do it.

OscarShiang:
I have force-pushed a commit for the following changes:
- enable UART0 FIFO to ensure that no any input will be lost.
- revise `demos/CMakeLists.txt` to avoid generating demos related to semihosting when semihosting feature is disabled.
- add a simple parser in `invalid_entry` to make debug easier

OscarShiang:
I revise the implementation of `strncpy` since I found that this implementation has a wrong behavior when the length of the string is equal to `n`.

I also remove the alignment after init_array to prevent calling the address that is not a part of the constructors.

OscarShiang:
I have added `strlen()`, `strncmp` and `strcmp` into the string library, since I observe that calling these functions with an unaligned `char` address will also cause data abort.

cmake/AddDemo.cmake
```diff
@@ -28,21 +28,28 @@ function(__add_demo NAME TEST_TYPE MAX_THREADS)
...
-  add_custom_command(TARGET debug_${NAME} POST_BUILD
-    COMMAND eval "${QEMU}${NAME} -s -S"
-    VERBATIM)
+    add_custom_command(TARGET debug_${NAME} POST_BUILD
+      COMMAND eval "${QEMU}${NAME} -s -S"
+      VERBATIM)
+  else()
+    # Generate kernel images
+    add_custom_command(TARGET ${NAME} POST_BUILD
+      COMMAND eval "${PREFIX}objcopy -O binary ${NAME} kernel8.img"
```
DavidSpickett: It may make more sense to skip the `run_` and `debug_` targets completely for raspi4. Instead add this objcopy command as part of the build step.

cmake/AddLoadable.cmake
DavidSpickett: This still applies.

demos/CMakeLists.txt
```diff
@@ -6,22 +6,26 @@ add_subdirectory(conditionvariables)
...
 add_subdirectory(yielding)
+
+if (SEMIHOSTING)
+  # These demos can only be executed when semihosting is enabled
+  add_subdirectory(file)
+  add_subdirectory(loadbinaries)
+  add_subdirectory(loadbinary)
+  add_subdirectory(loadpiebinary)
+  add_subdirectory(shell)
+endif()
```
DavidSpickett: Instead of doing this here, change the CMakeLists.txt of each demo. E.g.
```
-add_demo(file 6)
+if(NOT SEMIHOSTING)
+  add_demo(file 6)
+endif()
```

src/hw/raspi4/exception.c
DavidSpickett: Ah I see why you need this now. If there's a fault during early boot.

Still, including that in the comment is a good idea.

src/hw/raspi4/uart.c
```diff
@@ -0,0 +1,89 @@
...
+    /* convert newline to carrige return + newline */
+    if (*s == '\n')
+      uart_send('\r');
+    uart_send(*s++);
+  }
+}
+
+void uart_hex(unsigned int d) {
+  unsigned int n;
+  int c;
```
DavidSpickett: Declare the vars where used, so `for (int c = 28...` and `unsigned int n = (d >> c)...`.

src/hw/raspi4/uart.c
```diff
@@ -0,0 +1,89 @@
...
+  for (c = 28; c >= 0; c -= 4) {
+    // get highest tetrad
+    n = (d >> c) & 0xF;
+    // 0-9 => '0'-'9', 10-15 => 'A'-'F'
+    n += n > 9 ? 0x37 : 0x30;
```
DavidSpickett, line 84: 0x30 seems right, that's '0'. 0x37 is '7', shouldn't that side be 0x41 for 'A' ?

src/hw/raspi4/uart.c
```diff
@@ -0,0 +1,89 @@
...
+  /* read it and return */
+  char r = (char)(*UART0_DR);
+
+  /* convert carrige return to newline */
+  return r == '\r' ? '\n' : r;
+}
+
+void uart_puts(char* s) {
+  while (*s) {
+    /* convert newline to carrige return + newline */
```
DavidSpickett: Prefer `//` for single line comments.

include/port/raspi4/gpio.h
DavidSpickett: Are these used at all yet? Later you "map UART0 to GPIO pins" but that doesn't use these. Can this file be removed?

src/hw/raspi4/startup.s
```diff
@@ -0,0 +1,114 @@
...
+  // jump to C code, should not return
+  bl entry
+
+  // for failsafe, halt this core too
+  b on_other_cores
+
+  .macro ventry label
+    .align 7
+    b \label
+  .endm
```
DavidSpickett: Unindent this one level. (so it doesn't look like part of `at_el1`)

src/hw/raspi4/startup.s
```diff
@@ -0,0 +1,114 @@
...
+  mov sp, x1
+  
+  // jump to C code, should not return
+  bl entry
+
+  // for failsafe, halt this core too
+  b on_other_cores
+
+  .macro ventry label
+    .align 7
```
DavidSpickett: I think I understand this one. Based on https://ftp.gnu.org/old-gnu/Manuals/gas-2.9.1/html_chapter/as_7.html#SEC70

7 means 7 zeroes which would be 128 byte alignment.

Please add a comment explaining that. (or use `balign` which takes a number of bytes)

src/hw/raspi4/startup.s
DavidSpickett: So as above, 11 means 11 zeroes and 2^11 so 2048 byte alignment for the start of the tables.

Makes sense as the bottom bits of https://developer.arm.com/documentation/ddi0595/2020-12/AArch64-Registers/VBAR-EL1--Vector-Base-Address-Register--EL1- are fixed to zero. Please add a comment saying that.

DavidSpickett (changes requested):
I know this has been inactive a while (issues on my side) but if you're still around I've added some new feedback.

src/hw/raspi4/uart.c
OscarShiang: For 0xA, which is 10 in decimal. When we add 0x37, which is 55 in decimal, to it. 
It will become 10 + 55 = 65 (0x41 in hex) for 'A'.

And then for 0xF (15 in decimal). When we add 0x37 (55) to it.
It will be 15 + 55 = 70 (0x46 in hex) for 'F'.

include/port/raspi4/gpio.h
OscarShiang: Yes, we use part of these GPIO pins (like `GPFSEL1`) to set up the UART.

Should I remove those unused pins?

OscarShiang:
Hi, 
It has been a long time since the last contact. 
I have forced-pushed and updated the commit to fix the issues above.

demos/CMakeLists.txt
```diff
@@ -6,22 +6,22 @@ add_subdirectory(conditionvariables)
...
 add_subdirectory(stackcheck)
 add_subdirectory(threadlocalstorage)
 add_subdirectory(timer)
 add_subdirectory(trace)
 add_subdirectory(yielding)
+add_subdirectory(file)
+add_subdirectory(loadbinaries)
+add_subdirectory(loadbinary)
+add_subdirectory(loadpiebinary)
+add_subdirectory(shell)
```
DavidSpickett: Remove the changes to the order in this file.

include/port/raspi4/gpio.h
DavidSpickett: My bad I missed that, no leave them in and don't change this file.

src/hw/raspi4/startup.s
DavidSpickett: Now clear with comments, thanks.

src/hw/raspi4/uart.c
DavidSpickett: Got it, I was missing that the >10 side is offset by 10 to begin with.

src/hw/raspi4/libs/string.c
```diff
@@ -0,0 +1,90 @@
+/**
+ * We made our own implementation here because
+ * toolchain-provided implementation contains
+ * some instructions that has a strict rule to use.
+ *
```
DavidSpickett, line 5: FYI, I've been looking at building without the clib in general. (we build "freestanding" but add `-lc` to the link at the moment)

So don't change anything now but I might move this later into common/string.c and use it for all the ports. (compiled with `-mstrict-align` so that it still works for the rpi4)

src/hw/raspi4/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+  ${CMAKE_SOURCE_DIR}/src/hw/raspi4/yield.S
+  ${CMAKE_SOURCE_DIR}/src/hw/raspi4/uart.c
+  ${CMAKE_SOURCE_DIR}/src/hw/raspi4/mbox.c
+  ${CMAKE_SOURCE_DIR}/src/hw/raspi4/exception.c
+  ${CMAKE_SOURCE_DIR}/src/hw/raspi4/libs/string.c
+)
+
+set( PREFIX       "aarch64-none-elf-" )
+# Don't generate NEON
+set( PLATFORM     "-mcpu=cortex-a72 -mgeneral-regs-only -mstrict-align" )
```
DavidSpickett, line 18: Comment here to highlight the use of strict align

src/hw/raspi4/libs/string.c
```diff
@@ -0,0 +1,90 @@
...
+  }
+  return dst;
+}
+
+void* memmove(void* dst, const void* src, size_t len) {
+  uint8_t* d = dst + len;
+  const uint8_t* s = src + len;
+  while (len-- > 0) {
+    *--d = *--s;
+  }
```
DavidSpickett, line 101: I think your memmove is missing the other 2 possible situations. Though it probably only gets called for the one you've implemented, which itself looks fine. Since the compiler's only gonna use it when memory might overlap.

My impression of reading the docs is that you need to handle:
* src == dst, where you don't copy anything
* src < dst, you copy backwards starting at index (N-1) in the src (you've got this one currently)
* src > dst, you copy forwards starting at index 0 in the src

DavidSpickett (changes requested):

OscarShiang:
I have updated the commit with the following changes:
- restore the order of CMakeLists in demos.
- append comments in raspi4 CMakeLists to illustrate the usage of `-mgeneral-regs-only` and `-mstrict-align`.
- eliminate the redundant space in raspi4's startup assembly.

DavidSpickett (approved):
This looks great, thanks so much for your hard work!
(the exception diagnostics I might pull across into the virt port too, very useful)

I'm going to get this in building in the CI. (without tests of course)

Could you add a copy of https://hackmd.io/@oscarshiang/amt_setup to the /docs folder in a follow up change?

OscarShiang:
> Could you add a copy of https://hackmd.io/@oscarshiang/amt_setup to the /docs folder in a follow up change?

Sure, I have created a new PR for that in #30.

PR #29, opened 2021-07-11
Exclude CMakeFiles from LIT testing list

OscarShiang:
In this commit, I exclude the directory `CMakeFiles` from adding into the testing list
to prevent LIT from seeing it as a demo.


DavidSpickett:
Thanks! (CI is building out of tree that's why it didn't come up there)

